PR #1, opened 2016-10-11
Updated framework requirements

WilcoFiers:


ACT-reqs.md
```diff
@@ -0,0 +1,84 @@
...
+
+The ACT Framework should ensure that the rules are readable by both tool developers and less-technical users. For instance, the ACT Framework may discourage the use of pseudo code in favor of plain English.
+
+### 2.2 Ensure Justification
+
+The ACT Framework should ensure that ACT Rules are well grounded in the requirements they test. It is insufficient for an ACT Rule to merely claim something violates a requirement. The ACT Framework has to ensure that ACT Rules document why its claims are a valid interpretation of the requirement.
+
+### 2.3 Ensure Consistency
+
+The ACT Framework has to ensure that ACT Rules can be implemented in a way that provides consistent results. An ACT Rule must include a way to test the implementation of that rule.
```
rdeltour: What do you mean by "include a way to test the implementation"?

Also, I wouldn't use *must*, which makes it sound like speccing rules already.

ACT-reqs.md
```diff
@@ -0,0 +1,84 @@
...
+
+### 2.4 Ensure Measurable Accuracy
+
```
rdeltour: *measurable* sounds confusing to me (maybe it's just me, as a non-native speaker), as it can have both a strict meaning (evaluate a quantity/size/amount, with units) or a laxer one (assess the importance/value).
I'm not sure the prose below clarifies that (to be discussed).

ACT-reqs.md
```diff
@@ -0,0 +1,84 @@
...
+The ACT Framework ensures that ACT Rules can be tested for real world accuracy. False positives happen in every rule, as there are endless different ways to use web technologies. The ACT Framework ensures that common false positives can be identified and resolved. This also gives insight into how often a rule may give incorrect results.
+
+### 2.5 Ensure Evolution
+
+As web technologies change, the rules for testing them should change as well. Because of this it is important that ACT Framework outlines a way to deal with that. Rules should be dated and versioned, so that users of rules can understand that a rule needs to be updated. A possible approach to this could be semantic versioning, although there are many other possibilities.
+
+### 2.6 Existing Rulesets can be Transformed to ACT Rules
+
+The ACT Framework must ensure that existing rulesets can be updated to meet the ACT Framework. Existing work in this field should be the bases for the ACT Framework to ease the adoption. At least one ruleset must be adopted to the ACT Framework during the development process.
```
rdeltour: > At least one ruleset must be adopted 

Is it the Rec track requirement? Sounds more like a process req than a req on the framework itself.

ACT-reqs.md
```diff
@@ -0,0 +1,84 @@
...
+Status of This Document
+-----------------------
+
+This is the first draft of this document, it has not yet had the full review of the ACT Taskforce. It is therefore likely to change significantly.
+
+
+1. Introduction
+---------------
+
+There are many accessibility test tools (ATTs). These can be fully automatic, or can prompt a user for specific actions, in order to discover accessibility issues of some digital product. Each of these tools will run one, or a serie of tests to come to some result. These test procedures may be designed to look for violations of WCAG 2.0, or some other accessibility standard, and for some that are highly customizable they can even be used in testing an organization's own accessibility policy. The test procedures used by ATT are often called 'rules'.
```
rdeltour: The application scope of ACT is mentioned here ("WCAG 2.0, or some other accessibility standard"), but not in requirements.
If possible, there should be a requirement about this IMO, to refine which kind of technologies can be tested by an ACT rule.

maryjom: "series" is misspelled above as "serie"

ACT-reqs.md
maryjom: In 2.3, I would use the word "produces" rather than "provides".

ACT-reqs.md
maryjom: 'bases' should be 'basis'
I'm confused by the last statement, "At least one ruleset must be adopted to the ACT Framework during the development process."  If the ACT framework is to be a Recommendation, doesn't it require two implementations?

ACT-reqs.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. Introduction
+---------------
+
+The Accessibility Conformance Testing (ACT) Framework will be a recommendation on how to write rules for accessibility conformance testing. For an introduction on Accessibility Conformance Testing, read [What is ACT](https://www.w3.org/WAI/GL/task-forces/conformance-testing/wiki/ACT_Overview_-_What_is_ACT).
+
+The ACT Framework is developed by the [ACT Taskforce](https://www.w3.org/WAI/GL/task-forces/conformance-testing/), lead by the [WCAG Working Group](https://www.w3.org/WAI/GL/). The ACT Framework will standardize those aspects all rules have in common that ensure them to be reliable, implementable by different ATTs and transparent for the users of this tool. By standardizing this, the ACT Taskforce aims to promote a common understanding of how accessibility can be tested for certain technologies.
+
+Development and implementation of ACT Rules are outside the scope of the ACT Framework. This work is currently done within the [Auto-WCAG Community Group](https://www.w3.org/community/auto-wcag/). The ACT Taskforce support the Auto-WCAG Community group in the development of the first set of ACT Rules.
```
maryjom: Editorial: Last sentence should say "supports" or "will support"

ACT-reqs.md
a11ydoer: sorry for picking on minor thing. "These can be **fully** automatic, or can prompt a user for specific actions, in order to discover accessibility issues of some digital product." Do we need "fully" for this? Is there anything **fully** automatic testing?

How we deal with acronym in the spec such as ATT?  Some other w3c doc I have seen has a link to the terminology including the definition.

PR #10, opened 2016-11-17
Made changes to include the iBM rules description

cpandhi:
Added a column to include the IBM rules information.


cpandhi:
Starting over.

PR #24, opened 2016-12-20
Update Selector definition

moekraft:
The consensus from the 12/14/16 meeting was to simplify this definition
to use plainer text. As well, Alistair asks to be careful about the use
of element vs. node.

Some simplified text suggestions:
"selectors are patterns used to select the element(s) you want to work
with." - Alistair
From CSS 3 Selectors, Selectors are patterns that match against elements
in a tree, and as such form one of several technologies that can be used
to select nodes in an XML document. Selectors have been optimized for
use with HTML and XML, and are designed to be usable in
performance-critical code.
From w3 Schools: In CSS, selectors are patterns used to select the
element(s) you want to style.

### DIFF
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -63,7 +63,7 @@ ACT Test Procedure {#test-proc}
 Selector {#test-proc-selector}
 --------
 
-A selector is a pattern used to select element(s) in an HTML document or node(s) in the DOM tree. In an ACT Rule, a selector is applied to the entire input data to filter it into a set of elements or nodes to be evaluated against the test procedure.
+A selector is a pattern that is used as a condition to filter input data to be evaluated against the test procedure. For example, finding all  nodes in the DOM tree, or finding tags that are incorrectly closed in an HTML document.
 
 Selector syntax depends on the input type. When the input data is an HTML document or set of elements, the selector must be a CSS selector. When a formal selector syntax is not available for the input type, the selector may be described unambiguously in plain English.
 
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -63,7 +63,7 @@ ACT Test Procedure {#test-proc}
 Selector {#test-proc-selector}
 --------
 
-A selector is a boolean predicate that takes an element or component in the set represented in the input data and unambiguously tests whether the element or component matches the selector or not. In an ACT Rule, a selector is applied to the entire input data to filter it into a set of elements or components to be evaluated with the test procedure.
+A selector is a pattern used to select element(s) in an HTML document or node(s) in the DOM tree. In an ACT Rule, a selector is applied to the entire input data to filter it into a set of elements or nodes to be evaluated against the test procedure.
 
 Selector syntax depends on the input type. When the input data is an HTML document or set of elements, the selector must be a CSS selector. When a formal selector syntax is not available for the input type, the selector may be described unambiguously in plain English.
 


moekraft:
Not sure if this works for others but here is the update generated through rawgit and bikeshed applied. https://api.csswg.org/bikeshed/?url=https://cdn.rawgit.com/moekraft/wcag-act/Selector-definition/act-framework.bs#test-proc-selector

act-framework.bs
```diff
@@ -63,7 +63,7 @@ ACT Test Procedure {#test-proc}
 Selector {#test-proc-selector}
 --------
 
-A selector is a boolean predicate that takes an element or component in the set represented in the input data and unambiguously tests whether the element or component matches the selector or not. In an ACT Rule, a selector is applied to the entire input data to filter it into a set of elements or components to be evaluated with the test procedure.
+A selector is a pattern used to select element(s) in an HTML document or node(s) in the DOM tree. In an ACT Rule, a selector is applied to the entire input data to filter it into a set of elements or nodes to be evaluated against the test procedure.
```
WilcoFiers: We shouldn't limit ourselves to HTML document or DOM Tree. Instead. I'd suggest:

`A selector is a procedure for locating elements in a web page that will be tested by the ACT Rule. For example, finding all img nodes in the DOM tree, or finding tags that are incorrectly closed in a HTML document. A selector is applied to the entire input data, gather from it, a set of elements or nodes to be evaluated against the test procedure.`

rdeltour:
Some comments I had sent to @moekraft in an earlier email:

> - as I said in the call the expression "a pattern used to select something" sounds fuzzier than the previous wording to my ears. But I'm a developer and non-native speaker, so I'll just trust you all on this one :-)
> (I know this comes from CSS Level 3, but they're about to change that definition in Level 4).
> - When working in the previous wording,  I assumed that a selector wouldn't be limited to "select element(s) in an HTML document or node(s) in the DOM tree".  For instance, in digital publishing, the input data can be a collection of HTML documents. Maybe we need to get group consensus on how to define the input data first?
> - I don't think we need to distinguish "HTML document" or DOM tree. DOM is just one representation of an HTML document. I think the relevant distinction is between "static HTML" (as described in the file) vs "rendered/dynamic HTML" (as exposed in the browser DOM), but this distinction has to be made in the "Input data" section, not here in the selector section.
> - more of a nitpick: someone used "plain language" instead of "plain English" in another part of the doc, which I think would be a better wording here too.

act-framework.bs
rdeltour: I prefer @WilcoFiers' wording in that it's more general, but with still a couple nitpicks:

- I don't think a selector is a `procedure` (i.e. a series of actions), it's more a _declarative thing_ used in a procedure.
- strictly speaking, the selector isn't used to _locate_ elements, but in our case to _include_ them in the test set, or in other words to _filter_ the input data. The "location" information is secondary and not always relevant.

cpandhi: Prefer @WilcoFiers generalizing the wording and agree with @rdeltour. Here is what i suggest:
A selector is a structure or a pattern that is used as a condition to filter input data to be evaluated against the test procedure.

moekraft:
Hi all, 

Incorporating feedback from Wilco, Charu and Romain:

_A selector is a pattern that is used as a condition to filter input data to be evaluated against the test procedure. For example, finding all  nodes in the DOM tree, or finding tags that are incorrectly closed in an HTML document._ 

@charu, I don't think we should call a selector a structure. The structure would be the document or elements in the document. 

@rdeltour We strayed away from the CSS4 definition while we thought it was too technical. CSS selectors are generally called patterns. I agree a selector could select a set of web pages, however, would we still not need to apply a selector to each page before executing a test, e.g. if we were to check that each page has a title or a lang attribute?

maryjom:
I have a question on Moe's proposed update: Is the selector the pattern condition being used to determine if the input data is included in what content will be tested with the rule, or is it the actual data structure used to store the list of content items from the input data that matches the criteria?

rdeltour:
The new definition look good to me!

If I was to nitpick, I would say "finding tags that are incorrectly closed in an **serialized** HTML document", otherwise +1!

moekraft:
I have updated the Selector definition based off the feedback received from reviewing this pull request. Update has been approved to merge during 1/4/16 meeting.

PR #25, opened 2016-12-20
Update to Rule Description

moekraft:
Feedback from ACT meeting. Rule description should read:

Each ACT Rule MUST have a description that:

- Is in plain language;
- Explains the accessibility requirement being tested;
- Lists any limitations and assumptions for the test;
- Explains the failure conditions and their reasons.

### DIFF
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -7,6 +7,7 @@ Level: 3
 Status: w3c/FPWD
 Group: Accessibility Conformance Testing Task Force
 Editor: Wilco Fiers, Deque Systems
+Editor: Maureen Kraft, IBM Corp.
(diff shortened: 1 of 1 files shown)


cpandhi:
On re reviewing the whole section i feel the 3rd bullet is not needed in the rule description as we explicitly state all assumptions made by the rule to ensure accountability of the results in the Assumptions section.
 https://auto-wcag.github.io/auto-wcag/pages/design/rule-design.html#assumptions
Updated Rule description should read:
Each ACT Rule MUST have a description that: 

- Is in plain language; 

- Explains the accessibility requirement being tested; 

- Explains the failure conditions and their reasons.

PR #31, opened 2017-01-04
input types section draft

WilcoFiers:


act-framework.bs
```diff
@@ -54,8 +54,55 @@ ACT Input Data {#input}
 Test Input Types {#input-types}
 -------------------------------
 
-Editor note: Some rules may run only on the template, and others require the full page to be rendered on screen before it can run. This sections describe the different types of content that can be tested by ACT Rules, and how rules should deal with each of these.
+Web pages, publications and applications go through many different stages before it is rendered to the end user. For example, PHP may be used to put various pieces of content into a template. That template is then sent as an HTML text document to a web browser, which in term parses it and turns it into a DOM tree, before rendering it to the screen. At each of these stages, accessibility tests could be run. We can divide this into two stages of testing, black-box and white-box testing.
```
nitedog: Suggestion: "web pages, *including* publications and applications, ...".
Rationale: According to WCAG these are all types of web pages rather than different types of content.
Side note: Do we know what is mean by "publications"? Is there a definition somewhere?

moekraft: +1 to Shadi's comment. 
Some additional comments:

- Web pages are plural. "Web pages, including publications and applications, go through many different stages before **_they are_** rendered to the end user."

- At each of these stages, accessibility tests **_can_** be run. We can divide this into two stages of testing, black-box**_,_** and white-box testing.

- I think we need to provide some definitions here for DOM, black-box and white-box testing.


-

cpandhi: +1 to Shadi's and Moe's Comment.
publication Definition in the Cambridge English Dictionary
dictionary.cambridge.org/us/dictionary/english/publication
B2 [ U ] the act of making information or stories available to people in a printed or electronic form: The brochure will be ready for publication in September. ... B2 [ C ] a book, magazine, newspaper, or document in which information or stories are published: Our latest publication is a magazine for health enthusiasts.

maryjom: I think Moe's 2nd bullet with the additional comma is incorrect and makes it read like there are 2 stages of testing and black box and white box testing.  Suggest instead "...into two stages of testing: black box and white box testing."

Do we need to expand PHP (PHP: Hypertext Preprocessor) or is it a well-known acronym that needs no explanation.

act-framework.bs
```diff
@@ -54,8 +54,55 @@ ACT Input Data {#input}
...
+Example: The expand state of an element with `role=menu` can be toggled if the enter key is pressed when it has focus.
+
+### White-box testing
+
+In white box testing, the source files are tested instead of the output of the build process. The advantage of testing against source files is that it gives direct feedback to a developer, instead of only after a full page can be build. Additionally, it can provide information about which parts of elements are static, and which are dynamic.
+
+There are generally two approaches to building a web document. Either through templating, or through scripting. The source files for most (if not all) websites are either templates or scripts (or a mix of the two). This gives two new types of tests. Template testing and composition testing.
+
+Unlike black-box testing, in white-box testing, the Accessibility Test Tool implementing the ACT Rule must be developed specifically for the template engine or component technology. I.e. An ATT build to test HTML Handlebars templates should and can not be used to test Pug templates.
+
```
maryjom: An ATT 'built' to test HTML seems like the correct verb tense.

act-framework.bs
```diff
@@ -54,8 +54,55 @@ ACT Input Data {#input}
...
+#### Template Testing
+
+A template is document that has 'open' fields that are filled with pieces of content or other templates. E.g. an HTML template contains a head with a variable title, a predefined navigation, and a variable content area.
+
+Example: `<img>` tags with a variable `src`, *must not* have static `alt` value.
+
```
maryjom: Editorial: ... *must not* have a static 'alt' value.

act-framework.bs
```diff
@@ -54,8 +54,55 @@ ACT Input Data {#input}
...
+#### Composition Testing
+
+A composition is an class or component that extends other native elements or other compositions, to build a higher level component. E.g. a login form component, consists of a form, a few fields, and a label.
+
```
maryjom: Editorial: ...is a class...

act-framework.bs
```diff
@@ -54,8 +54,55 @@ ACT Input Data {#input}
...
+Example: Component properties starting with `aria-` *must* exist in the list of ARIA attributes.
```
maryjom: For ARIA, are we required to use WAI-ARIA due to registered trademark issues?

act-framework.bs
WilcoFiers: agreed

act-framework.bs
WilcoFiers: I agree with Shadi's suggest. "web pages, including publications and applications, ..."

As for the rest. I think maybe test stages isn't the right phrasing here. Are they two categories of test types?

moekraft: I agree with Mary Jo, however, I believe we should use a semi-colon before black-box and white-box testing, "into two stages of testing; black-box and white-box testing." Note: hyphens are commonly used.

moekraft: http://softwaretestingfundamentals.com/differences-between-black-box-testing-and-white-box-testing/

rdeltour: +1 to the previous comments.
I don't think we need to define "publications" here, as this is general, informative prose. If we say we need to define it, then we'd also need to define "application".

PR #32, opened 2017-01-04
Accessibility support data description

WilcoFiers:
Ready for review and possibly merger.

### DIFF
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -72,11 +72,13 @@ A list of `Relied upon accessibility features` is an optional part of ACT Rule. 
 - Attributes: Specific attributes that are to be tested. If there are none, this can be left blank.
 
 Example 1: Requires support for the `alt` attribute on an `img` element, existing within a link.
+
 ```
   a[ref] img[alt]
 ```
 
 Example 2: Requires support for role=menuitem on any type of element, existing within an element marked up as a menu
+
 ```
   *[role=menu] *[role=menuitem]
 ```
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -59,14 +59,27 @@ Editor note: Some rules may run only on the template, and others require the ful
 Accessibility Support Data {#input-acc-supp}
 --------------------------------------------
 
-Determining if content is accessible, depends partly on the assistive technologies that are used to access it. This is known as Accessibility Supported in WCAG 2.0. One accessibility technique can be relied upon on one site, given their users, but not on another. This depends entirely on the users of the site, and the accessibility support policy an organization sets for themselves.
+Determining if a web page is accessible, depends partly on the assistive technologies that are used to access it. This is known as [Accessibility Supported in WCAG 2.0][accsupp]. When testing a web page for accessibility, it is important to know which assistive technologies are to be used for this page.
 
-It would be impractical to write a new ACT Rule every time a new assistive technology is released. An ACT Rule should therefore be independent of assistive technologies. To do this, it must take accessibility support data, and use it in addition to the test page (or other test subject) to determine passing of failing of the rule.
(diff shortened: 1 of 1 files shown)


act-framework.bs
```diff
@@ -56,11 +56,16 @@ Test Input Types {#input-types}
...
 
-
 Accessibility Support Data {#input-acc-supp}
 --------------------------------------------
 
-Editor note: This section will describe data about the accessibility support of different assistive technologies should be used by rules to produce results. Where relevant, rules must be able to take in data about supported features in assistive technologies, and adjust results accordingly.
+Determining if content is accessible, depends partly on the assistive technologies that are used to access it. This is known as Accessibility Supported in WCAG 2.0. One accessibility technique can be relied upon on one site, given their users, but not on another. This depends entirely on the users of the site, and the accessibility support policy an organization sets for themselves.
+
+It would be impractical to write a new ACT Rule every time a new assistive technology is released. An ACT Rule should therefore be independent of assistive technologies. To do this, it must take accessibility support data, and use it in addition to the test page (or other test subject) to determine passing of failing of the rule.
+
```
maryjom: Editorial: ...passing **or** failing of the rule.

act-framework.bs
```diff
@@ -56,11 +56,16 @@ Test Input Types {#input-types}
 
 Editor note: Some rules may run only on the template, and others require the full page to be rendered on screen before it can run. This sections describe the different types of content that can be tested by ACT Rules, and how rules should deal with each of these.
 
-
 Accessibility Support Data {#input-acc-supp}
 --------------------------------------------
 
-Editor note: This section will describe data about the accessibility support of different assistive technologies should be used by rules to produce results. Where relevant, rules must be able to take in data about supported features in assistive technologies, and adjust results accordingly.
+Determining if content is accessible, depends partly on the assistive technologies that are used to access it. This is known as Accessibility Supported in WCAG 2.0. One accessibility technique can be relied upon on one site, given their users, but not on another. This depends entirely on the users of the site, and the accessibility support policy an organization sets for themselves.
+
```
maryjom: Editorial: Remove comma after the phrase 'Determining if content is accessible'

act-framework.bs
```diff
@@ -56,11 +56,16 @@ Test Input Types {#input-types}
...
+It would be impractical to write a new ACT Rule every time a new assistive technology is released. An ACT Rule should therefore be independent of assistive technologies. To do this, it must take accessibility support data, and use it in addition to the test page (or other test subject) to determine passing of failing of the rule.
+
+TODO: Describe what format the accessibility support data should be in.
+
+TODO: Describe how an ACT Rule should use the accessibility support data.  I'm thinking we could use a CSS selector style description like this: `input[type="text"][aria-describedby]` It would mean that anything that doesn't support aria-describedby, or aria-describedy on input or aria-describedby on input type="text" would not be accessibility supported.
```
nitedog: I think this is an interesting approach. It would potentially give us a notation for accessibility support in rules. However, for now I'm wondering how much of this should be decided in this requirements document rather than in the spec itself. Maybe put this as an example notation so that we can continue discussing it without holding up the completion of this requirements document?

act-framework.bs
moekraft: Not understanding if the rule is independent of Accessibility Support Data, why would the rule "must take accessibility support data, and use it in addition to the test page (or other test subject) to determine passing of failing of the rule."?

act-framework.bs
rdeltour: Editorial: add a link to the WCAG definition of Accessibility Supported?

act-framework.bs
rdeltour: Editorial: change "it must take" to "it has to take" (or is the use of an RFC2119 keyword intentional here?)

maryjom:
On [8 February 2015 survey](https://www.w3.org/2002/09/wbs/93339/ACTTF8Feb2017/).

PR #41, opened 2017-01-11
First draft Data format

WilcoFiers:


act-framework.bs
```diff
@@ -86,8 +86,34 @@ ACT Output Data {#output}
...
+Rules are tested in two steps. Firstly, the selector works out what elements / tags or other 'components' to test, these are known as the *Test Target*. Following this, each test target will be taken through the test procedure of the rule. This will give a *Test Result* for each *Test Target*.
 
+In addition to Test Target and Test Result, contextual information MUST be provided to indicate what web page, file or other test subject the rule was applied to, as well as an identifier of the rule itself. Running a rule will result in at least the following four pieces of information:
+
+- Test Criterion: The rule (or other test) that was used to generate the result
+- Subject: The web page or other component that was tested
+- Target: An element / component contained within the subject, that was selected for testing
+- Result: The finding as determined by the test, either: Pass, Fail or Unknown
+
```
maryjom: Target bullet, suggest saying "or" rather than '/' in the sentence.

act-framework.bs
```diff
@@ -86,8 +86,34 @@ ACT Output Data {#output}
...
+### Test Target
+
+How to describe the test target depends on whether or not the test subject is parsed or not. When testing against the DOM tree, CSS and XPath are usually the way elements are located. Where as for files the most common method is to indicate line number and character number. For the purpose of validating
+
```
maryjom: Seems like there's an incomplete sentence "For the purpose of validating"

act-framework.bs
```diff
@@ -86,8 +86,34 @@ ACT Output Data {#output}
...
+- Text Files: Express the location of the target with two properties: Line number, and Character number.
+
+### Test Result
+
+A rule must result in one of the following three values:
+
+- Pass: The Test Target passed the test
+- Fail: The Test Target failed the test
+- CantTell: It is unclear whether or not the Test Target passes of rails the test
+
```
maryjom: In the earlier part of the text "CantTell" is "Unknown".  Pick one term.

act-framework.bs
rdeltour: how is the Target identified? there can be several ways: DOM node, unique CSS selector, XPath, etc. Does it need to be specified here?

act-framework.bs
```diff
@@ -86,8 +86,34 @@ ACT Output Data {#output}
...
+- Result: The finding as determined by the test, either: Pass, Fail or Unknown
+
+### Test Target
+
+How to describe the test target depends on whether or not the test subject is parsed or not. When testing against the DOM tree, CSS and XPath are usually the way elements are located. Where as for files the most common method is to indicate line number and character number. For the purpose of validating
+
+- DOM Tree: Express the target as either a CSS selector targeting a single node on the page, or an XPath selector.
+
+- Text Files: Express the location of the target with two properties: Line number, and Character number.
+
```
rdeltour: what is "Character number"? is it the column number or is it a character count for a text node?
Do we envision rules that apply to a range of characters?

act-framework.bs
moekraft: Editorial. Add an article before Test Target in the first sentence second paragraph. "In addition to the Test Target and Test Result"

act-framework.bs
WilcoFiers: I'll look at how this was done in EARL.

act-framework.bs
WilcoFiers: Should be CantTell

WilcoFiers:
Made a bunch of updates after the feedback from earlier. I think we are ready for another review.

act-framework.bs
```diff
@@ -129,11 +128,41 @@ Editor note: This section describes how rules are broken down into one or more t
...
 ACT Data Format {#output-format}
 --------------------------------
 
-Editor note: This section describes the required properties of the data output by a rule. Certain parts must be standardized to enable aggregation of results produced by different accessibility test tools. Additionally, standardizing parts of the output format is required for validating the implementation (see below). Data in the output format has to be accessible.
+Rules are tested in two steps. Firstly, the selector is applied to the web page or other test subject. This results in a list of *Selected item* (elements, tags or other 'components') to test. Following this, each *selected item* will be taken through the test procedure of the rule. This will give the *outcome* for each *selected item*. For contextual information, the output data must also include *test subject* and the *rule identifier*.
```
cpandhi: +1 to the question

WilcoFiers: Initially I had anticipated input data was going to include AT data, but we've since pulled that and made that stuff static, so yes, it's the same. I don't particularly like either term. We'll need to have a discussion about what we do with EARL terminology.

WilcoFiers: Resolution from today's meeting: Rename "input type" to "test subject type"

PR #42, opened 2017-01-11
Update rule description

cpandhi:
updates to the rule description to include scope and expand on the bullets are ready for review.

### DIFF
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -44,8 +44,8 @@ Each ACT Rule MUST have a description that:
 
 * Is in plain language and provides a brief description of what the rule does.
 * Explains the accessibility requirement being tested such as the WCAG SC or the technique the rule maps to; For example WCAG 2.0 Technique H67.
-* Lists any limitations and assumptions for the test, the test environment, technologies being used or the subject being tested;
-* The scope or the context of the rule; For example: A rule for 1.4.1: Use of Color has to make an assumption with CSS-properties are used to make a link visually evident. Typically something like background, border, color, font, or text-decoration.
+* Lists any limitations and assumptions for the test, the test environment, technologies being used or the subject being tested; For example: A rule for 1.4.1: Use of Color has to make an assumption with CSS-properties are used to make a link visually evident. Typically something like background, border, color, font, or text-decoration.
+* The scope or the context of the rule; 
 * Explains the failure conditions and their reasons. For example the rule triggers on all images on the page that has no alt attribute defined.
 
 
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -44,8 +44,8 @@ Each ACT Rule MUST have a description that:
 
 * Is in plain language and provides a brief description of what the rule does.
 * Explains the accessibility requirement being tested such as the WCAG SC or the technique the rule maps to; For example WCAG 2.0 Technique H67.
-* Lists any limitations and assumptions for the test, the test environment, technologies being used or the subject being tested; For example: A rule for 1.4.1: Use of Color has to make an assumption with CSS-properties are used to make a link visually evident. Typically something like background, border, color, font, or text-decoration.
-* The scope or the context of the rule; 
+* Lists any limitations and assumptions for the test, the test environment, technologies being used or the subject being tested; For example: A rule for 1.4.1: Use of Color has to make an assumption that CSS properties are used to make a link visually evident. Typical examples are: background, border, color, font, or text decoration.
+* Provides the scope or the context of the rule; 
 * Explains the failure conditions and their reasons. For example the rule triggers on all images on the page that has no alt attribute defined.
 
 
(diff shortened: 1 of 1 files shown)


act-framework.bs
```diff
@@ -42,10 +42,11 @@ Rule Description {#structure-description}
 
 Each ACT Rule MUST have a description that:
 
-* Is in plain language;
-* Explains the accessibility requirement being tested;
-* Lists any limitations and assumptions for the test;
-* Explains the failure conditions and their reasons.
+* Is in plain language and provides a brief description of what the rule does.
+* Explains the accessibility requirement being tested such as the WCAG SC or the technique the rule maps to; For example WCAG 2.0 Technique H67.
+* Lists any limitations and assumptions for the test, the test environment, technologies being used or the subject being tested; For example: A rule for 1.4.1: Use of Color has to make an assumption with CSS-properties are used to make a link visually evident. Typically something like background, border, color, font, or text-decoration.
```
maryjom: Editorial: '...has to make an assumption **that** CSS properties', don't hyphenate CSS properties, change 'with' to 'that'.  The last sentence is missing a verb, so suggest saying 'Typical examples are: background, border, color, font, or text decoration'.  Text decoration should not be hyphenated.

act-framework.bs
```diff
@@ -42,10 +42,11 @@ Rule Description {#structure-description}
...
+* The scope or the context of the rule;
```
maryjom: Need a verb for this to be consistent.  'Contains the scope or the context of the rule;'

moekraft:
This duplicates much of the Rule Outline. The Rule Description should be part of the Rule Outline and in plain language describe what the rule does..

moekraft:
See PR #43

act-framework.bs
```diff
@@ -42,10 +42,11 @@ Rule Description {#structure-description}
...
 
-* Is in plain language;
-* Explains the accessibility requirement being tested;
-* Lists any limitations and assumptions for the test;
-* Explains the failure conditions and their reasons.
+* Is in plain language and provides a brief description of what the rule does.
+* Explains the accessibility requirement being tested such as the WCAG SC or the technique the rule maps to; For example WCAG 2.0 Technique H67.
+* Lists any limitations and assumptions for the test, the test environment, technologies being used or the subject being tested; For example: A rule for 1.4.1: Use of Color has to make an assumption that CSS properties are used to make a link visually evident. Typical examples are: background, border, color, font, or text decoration.
+* Provides the scope or the context of the rule; 
+* Explains the failure conditions and their reasons. For example the rule triggers on all images on the page that has no alt attribute defined.
```
rdeltour: > the rule triggers o all the images on the page that has no alt attribute defined

do it mean the rule is applied to all such images, or returns a `fail` to all such images? (in other words, is it an example of the rule scope and context, or of the rule failure condition?)

moekraft: This duplicates much of the Rule Outline. The Rule Description should be part of the Rule Outline and in plain language describe what the rule does..

cpandhi:
Have pushed changes based on review.

act-framework.bs
```diff
@@ -42,10 +42,8 @@ Rule Description {#structure-description}
 
 Each ACT Rule MUST have a description that:
 
-* Is in plain language;
-* Explains the accessibility requirement being tested;
-* Lists any limitations and assumptions for the test;
-* Explains the failure conditions and their reasons.
+* Is in plain language and provides a brief description of what the rule does.
+* Explains the accessibility requirement being tested such as the WCAG SC or the technique the rule maps to; For example WCAG 2.0 Technique H67.
```
nitedog: If I recall correctly, a rule always applies to at least one success criterion, and may map to any number of techniques? If so, should this be separated into two points:
 - Lists the accessibility requirement being tested, such as the WCAG Success Criterion
 - Refers to any related information, such as the WCAG Techniques it maps to

act-framework.bs
```diff
@@ -42,10 +42,8 @@ Rule Description {#structure-description}
 
 Each ACT Rule MUST have a description that:
 
-* Is in plain language;
-* Explains the accessibility requirement being tested;
-* Lists any limitations and assumptions for the test;
-* Explains the failure conditions and their reasons.
+* Is in plain language and provides a brief description of what the rule does.
```
rdeltour: "a rule must have a description that provides a brief description of what the rule does" sounds a bit tautological 😉.
I'd suggest "provides a summary" or "provides a short explanation" instead.

act-framework.bs
cpandhi: Good point Shadi, i will make the suggested changes. Thx.

act-framework.bs
cpandhi: Like short explanation, will make the change.

cpandhi:
Have pushed updates based on review comments.

PR #43, opened 2017-01-18
Rule outline

moekraft:
This update provides a proposal for the Rule outline. 

Trying not to be too prescriptive here but are we inherently being prescriptive by generating a rule outline? 

Moe

### DIFF
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -35,7 +35,13 @@ ACT Rule Structure {#structure}
 Rule Outline {#structure-outline}
 -----------------------
 
-Editor note: This sections gives a broad outline of parts make up an ACT Rule. We'll go into further detail in sections below. At this point it is important to show the reader the big picture
+A rule MUST provide the following items written in plain language:
+* A descriptive title 
+* A unique identifier
+* A rule description
+* Associated criteria, standard(s) or requirement(s) (Note: Auto-WCAG calls this Background)
+* Assumptions, if any
+* Test procedure
 
 Rule Description {#structure-description}
 ----------------------------------------
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -36,10 +36,12 @@ Rule Outline {#structure-outline}
 -----------------------
 
 A rule MUST provide the following items written in plain language:
+
 * A descriptive title 
 * A unique identifier
 * A rule description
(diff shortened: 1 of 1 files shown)


act-framework.bs
```diff
@@ -35,7 +35,13 @@ ACT Rule Structure {#structure}
 Rule Outline {#structure-outline}
 -----------------------
 
-Editor note: This sections gives a broad outline of parts make up an ACT Rule. We'll go into further detail in sections below. At this point it is important to show the reader the big picture
+A rule MUST provide the following items written in plain language:
+* A descriptive title
```
WilcoFiers: Editorial: Leave a blank line between this paragraph and the list.

cpandhi: Should we add, pass / fail criteria to the rule outline or do we expect that to be included in the test procedure.

rdeltour: @cpandhi good idea (as far as I can see): the possible outcomes may differ from one rule to another, so could very well belong to the rule description

act-framework.bs
```diff
@@ -35,7 +35,13 @@ ACT Rule Structure {#structure}
 Rule Outline {#structure-outline}
 -----------------------
 
-Editor note: This sections gives a broad outline of parts make up an ACT Rule. We'll go into further detail in sections below. At this point it is important to show the reader the big picture
+A rule MUST provide the following items written in plain language:
+* A descriptive title 
+* A unique identifier
+* A rule description
+* Associated criteria, standard(s) or requirement(s) (Note: Auto-WCAG calls this Background)
```
rdeltour: the item above (associated criteria, etc) describe the rationale for the rule.
Another aspect are pointers to remediation techniques (for instance to WCAG Techniques), would that belong to another item?

act-framework.bs
moekraft: @charu pass/fail criteria would be part of the test cases. I'm using Auto WCAG as the inspiration.

act-framework.bs
moekraft: @rdeltour Correct. I think the techniques should fall under the test cases.

moekraft: @rdeltour During the meeting it was agreed to shorten Associated criteria, standards or requirements to simply Accessibility requirement(s) and to add an item for Related techniques.

PR #44, opened 2017-01-25
Update input type

WilcoFiers:
- Remove white/blackbox testing references
- Renamed a few of the input types for clearity

### DIFF
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -53,51 +53,44 @@ ACT Input Data {#input}
 
 Test Input Types {#input-types}
 -------------------------------
-Web pages, including publications and applications go through many different stages before it is rendered to the end user. For example, PHP may be used to put various pieces of content into a template. That template is then sent as an HTML text document to a web browser, which in term parses it and turns it into a DOM tree, before rendering it to the screen. At each of these stages, accessibility tests could be run. We can divide this into two stages of testing, black-box and white-box testing.
+Web pages, including publications and applications go through many different stages before it is rendered to the end user. For example, PHP may be used to put various pieces of content into a template. That template is then sent as an HTML text document to a web browser, which in term parses it and turns it into a DOM tree, before rendering it to the screen. At each of these stages, accessibility tests could be run. It is therefore important to specify the *type of input* that an ACT Rule expects.
 
-### Black-box testing
+The following input types are common in accessibility testing:
 
-In black box testing the final product is tested, without looking at the code that generated it. To do black-box testing, the tester does not need access to the source file(s), which usually makes black-box testing easier to set up.
+- HTTP Response
+- DOM Tree
+- Rendered Page
+- (Web)Driver
+- Template
+- Script
 
-There are different types of black-box testing:
-- File testing
-- DOM tree testing
-- Browser testing
-- Driver testing
+Other types of input may be possible. In those cases the ACT Rule *must* include a detailed description of the input type, or a reference to the that description.
 
-#### File Testing
+### HTTP Response Testing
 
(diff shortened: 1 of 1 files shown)


act-framework.bs
```diff
@@ -53,51 +53,45 @@ ACT Input Data {#input}
 
 Test Input Types {#input-types}
 -------------------------------
-Web pages, including publications and applications go through many different stages before it is rendered to the end user. For example, PHP may be used to put various pieces of content into a template. That template is then sent as an HTML text document to a web browser, which in term parses it and turns it into a DOM tree, before rendering it to the screen. At each of these stages, accessibility tests could be run. We can divide this into two stages of testing, black-box and white-box testing.
 
-### Black-box testing
+Web pages, including publications and applications go through many different stages before it is rendered to the end user. For example, PHP may be used to put various pieces of content into a template. That template is then sent as an HTML text document to a web browser, which in term parses it and turns it into a DOM tree, before rendering it to the screen. At each of these stages, accessibility tests could be run. It is therefore important to specify the *type of input* that an ACT Rule expects.
```
maryjom: Editorial: FIrst sentence, since "Web pages" is plural, "it is" should instead be "they are".

act-framework.bs
```diff
@@ -53,51 +53,45 @@ ACT Input Data {#input}
...
 
-There are different types of black-box testing:
-- File testing
-- DOM tree testing
-- Browser testing
-- Driver testing
+- HTTP Response
+- DOM Tree
+- Rendered Page
+- (Web)Driver
```
rdeltour: What is meant by Driver here?
Typically web drivers would provide access to a DOM or rendered page, so it would refer to the bullets above. Or is it meant to describe actionable drivers (where you can emulate user actions)?

rdeltour: It seems to me that this is not really of the same nature as the 3 bullets above. 
The previous items describe **content** while a Web Driver is more an API. Eventually, the driver returns a view of the content (typically a DOM tree), which is what is tested.

Maybe be "Actionable page" better describes what is meant here?

act-framework.bs
```diff
@@ -53,51 +53,45 @@ ACT Input Data {#input}
...
+- Template
```
rdeltour: Just a question, are templates just listed for sake of completeness, or are they actually in scope for ACT?

act-framework.bs
```diff
@@ -53,51 +53,45 @@ ACT Input Data {#input}
...
+### Template Testing
 
 A template is document that has 'open' fields that are filled with pieces of content or other templates. E.g. an HTML template contains a head with a variable title, a predefined navigation, and a variable content area.
 
 Example: `<img>` tags with a variable `src`, *must not* have a static `alt` value.
 
-#### Composition Testing
+### Script Testing
 
 A composition is a class or component that extends other native elements or other compositions, to build a higher level component. E.g. a login form component, consists of a form, a few fields, and a label.
```
maryjom: This needs some editing, as the item is "Script Testing" but the definition calls it "composition".

act-framework.bs
WilcoFiers: I'm not sure why the DOM Tree would be any less of an API then Webdriver. I suppose it should be '(Web)Driver page, for the sake of consistency.

act-framework.bs
WilcoFiers: They are in scope, although I'd say it's more there for the sake of completeness, and possibly to give someone the idea that this sort of thing can be done. I don't know of anyone who is doing this, but someone might be.

act-framework.bs
WilcoFiers: Quite right. I suggest the following:

A script test is a test in which, through static code analyses, components and other elements are tested. 

Example: Component properties starting with `aria-` *must* exist in the list of WAI-ARIA attributes.

PR #46, opened 2017-02-06
Add first draft aggregation section

WilcoFiers:


act-framework.bs
```diff
@@ -138,8 +138,86 @@ Editor note: This section describes the required properties of the data output b
 Rule Aggregation {#output-aggregation}
 --------------------------------------
 
-Editor note: In this section we describe how the data that is returned from a rule can be combined to give a higher level view of the conformance to accessibility requirements. Rules provide very low level information, this is valuable for people working at that level, but managing accessibility of products as a whole requires a higher level understanding of the accessibility.
-
+As describe in section [ACT Data Format](#output-format) a rule will return a list of results, each of which contain 1) an outcome (passed, failed or cantTell), 2) the selected item, 3) the Rule ID and 4) the test subject. We can create a new result, that combines all the results of that rule, to create provide a single result for that rule.
```
nitedog: This last sentence does not make sense grammatically.

act-framework.bs
```diff
@@ -138,8 +138,86 @@ Editor note: This section describes the required properties of the data output b
 Rule Aggregation {#output-aggregation}
```
nitedog: Is this really in our purview to define how individual test results should be aggregated? Isn't the scope to just define the rule format and validation/benchmarking?

act-framework.bs
```diff
@@ -138,8 +138,86 @@ Editor note: This section describes the required properties of the data output b
...
-
+As describe in section [ACT Data Format](#output-format) a rule will return a list of results, each of which contain 1) an outcome (passed, failed or cantTell), 2) the selected item, 3) the Rule ID and 4) the test subject. We can create a new result, that combines all the results of that rule, to create provide a single result for that rule.
+
+Individual cases are useful for remediation, but often what is needed is a higher level view of accessibility. For instance, when reporting accessibility to higher management, all they are interested in is which websites aren't meeting the requirements.
+
+An aggregated result takes puts the original data as it's source data. Aggregated results do not have a pointer. The outcome of the aggregated result, is the worst outcome on the source data. (E.g. If there is one Failed, and ten Passed, the aggregated outcome is Failed)
```
rdeltour: Editorial:
- s/takes puts/takes/
- s/it's/its/

act-framework.bs
WilcoFiers: We need to define how aggregation should be done, in order to do benchmarking correctly. In benchmarking, we want to take results from accessibility experts, and compare them to results from ATTs. The problem is that experts don't report with the same level of detail that ATTs do. To compare data from one with the other, they need to be at the same level of specificity. If an expert tester reports pass/fail on a criterion level, per page, we'll need to do the same with the results from an ATT before we can compare, hence aggregation.

act-framework.bs
WilcoFiers: should be "that combines all the results of that rule, to create a single result for that rule."

maryjom:
On [8 Feb. 2017 survey](https://www.w3.org/Talks/Tools/Slidy2/#(1))

PR #47, opened 2017-02-06
First draft benchmark

WilcoFiers:


act-framework.bs
```diff
@@ -159,8 +159,17 @@ Editor note: This section describes the requirements of tests that have to be cr
...
+The web is ever changing, and technologies are used in such diverse and creative ways that it is impossible to predict in advance, all the ways that accessibility issues can occur and all the ways they can be solved for. When writing ACT Rules, it is almost inevitable that exceptions will be overlooked during the design of a rule, or that new technologies will emerge that introduce new exceptions.
 
+This makes it important to be able to regularly test if the rule has the accuracy that is expected of it. This can be done by benchmark testing. In benchmark testing, the accuracy of a rule is measured, by comparing it's results to those obtained through accessibility expert testing.
+
+There are several ways this can be done. For instance, accessibility test tools can implement a feature which lets users indicate that that a result is in error, or pages that for which accessibility results are known, can be tested using ATT, and the results are compared.
+
+The accuracy of a rule is the average between the false positives and false positives, which are in tern calculated as follows:
```
nitedog: Typo: "in tern" -> "in turn"

act-framework.bs
```diff
@@ -159,8 +159,17 @@ Editor note: This section describes the requirements of tests that have to be cr
...
+The web is ever changing, and technologies are used in such diverse and creative ways that it is impossible to predict in advance, all the ways that accessibility issues can occur and all the ways they can be solved for. When writing ACT Rules, it is almost inevitable that exceptions will be overlooked during the design of a rule, or that new technologies will emerge that introduce new exceptions.
 
+This makes it important to be able to regularly test if the rule has the accuracy that is expected of it. This can be done by benchmark testing. In benchmark testing, the accuracy of a rule is measured, by comparing it's results to those obtained through accessibility expert testing.
+
+There are several ways this can be done. For instance, accessibility test tools can implement a feature which lets users indicate that that a result is in error, or pages that for which accessibility results are known, can be tested using ATT, and the results are compared.
```
nitedog: So do we have two "validation" steps?
1. When publishing a rule, I guess there is some sort of checking
2. On-going benchmarking to make sure rules are still accurate

rdeltour: Editorial:

- s/the accuracy of a rule is measured, by comparing it's results/the 
- I'd split the examples in the last paragraph in 2 distinct sentences, for better clarity.

act-framework.bs
rdeltour: Editorial: s/false positives and false positives/false positives and false negatives/

rdeltour: I would say false positives are worse than false negatives, should this have an impact on the evaluation of the accuracy of the rule? (in other words, should we add some weights to the calculation?). Not sure about this one...

act-framework.bs
cpandhi: +1 to Shadi's comment on 2 steps.

WilcoFiers: @nitedog It's not really 2 steps, it's an iterative process that is repeated ever so often.

act-framework.bs
WilcoFiers: @rdeltour I think from a dev perspective false positives are worse, I think from a user perspective false negative are worse. I don't think we should put a value to that.

moekraft: "There are several ways this can be done. For instance, accessibility test tools can implement a feature which lets users indicate that that a result is in error, or pages that for which accessibility results are known, can be tested using ATT, and the results are compared."

- Second that in the last sentence
- Need ATT acronym definition

moekraft: Hi Wilco, After today's conversation, I thought of a real life example. In support of WCAG Sufficient Technique [H44](https://www.w3.org/TR/WCAG20-TECHS/H44.html), we had a rule to check for form labels. However, because at the time of writing the rule, some screen readers did not support implicit labels, see [UA note](https://www.w3.org/WAI/WCAG20/Techniques/ua-notes/html#H44), the rule would only pass if explicit labels were used. Recently, we received feedback from our development teams that implicit labels, wrapping input elements inside a <label> element, works just fine and are read by the screen reader. We have updated our rule to support both implicit and explicit labels while is supported by both HTML and XHTML specifications to associate text labels with form controls and meets the requirements of the related WCAG SC.

moekraft: As we noted previously, our rules should not be based on whether or not a screen reader supports the specification.

maryjom:
On [8 Feb. 2017 survey](https://www.w3.org/Talks/Tools/Slidy2/#(1))

PR #51, opened 2017-02-06
Mangaging exceptions

cpandhi:
Add separate for accessibility requirements and limitations / assumptions / exceptions for the test under rule description.

### DIFF
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -47,6 +47,13 @@ Each ACT Rule MUST have a description that:
 * Lists any limitations and assumptions for the test;
 * Explains the failure conditions and their reasons.
 
+Accessibility Requirements {#structure-accessibility-requirements}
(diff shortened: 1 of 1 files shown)


maryjom:
On [15 Feb. 2017 survey](https://www.w3.org/2002/09/wbs/93339/ACTTF15Feb2017/).

moekraft:
If we are going to have separate sections for Accessibility requirements and Limitations, Assumptions and Exceptions, then these should be removed from the Rule Description section. Should we also add a section for Related techniques? 

[Rule Outline](https://w3c.github.io/wcag-act/act-framework.html#structure-outline) currently has the following items:
A rule MUST provide the following items written in plain language:

- A descriptive title
- A unique identifier
- A rule description
- Accessibility requirement(s)
- Related techniques
- Assumptions, if any
- Test procedure

Issue #29, opened 2016-12-21
Draft Section 5.2 Test Cases


moekraft:
Editor note: This section describes how rules are broken down into one or more test cases. Each test case gives some result that, when combined, provides the outcome of the rule. Additionally, this section describes how rule authors should write test cases and the mechanism of combining their outcomes.

moekraft:
Original draft in pull request #54.

moekraft:
Pull request #54 has been merged and closed.

PR #54, opened 2017-02-15
Test cases - Issue #29

moekraft:
Original draft of Test cases section per issue #29.

### DIFF
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -131,12 +131,14 @@ Selector syntax depends on the input type. When the input data is an HTML docume
 Test Cases {#test-proc-cases}
 ---------------------------------------
 
-Each rule contains one or more testcases written in plain language. The testcases provide the following:
+Each rule contains one or more testcases written in plain language. Each testcase MUST provide the following:
+
 *Testcase mode identifying whether testcase steps are automated, semi-automated or require manual testing.
 *Testcase steps which clearly outline the steps necessary to reach the result.
-*Test case result of whether testcase passed or failed.
-**Failing test results provide a description of the failure.
-**Each test case result that, when combined, provides the outcome of the rule. 
+*Test case result indicating whether testcase passed or failed.
+**Description of error causing any failing result.
+
+Test case results are combined with other test case results to provide a cumulative outcome, pass or fail, of the rule. 
 
 
 
--- a/act-framework.bs
+++ b/act-framework.bs
@@ -131,7 +131,13 @@ Selector syntax depends on the input type. When the input data is an HTML docume
 Test Cases {#test-proc-cases}
 ---------------------------------------
 
-Editor note: This section describes how rules are broken down into one or more test cases. Each test case gives some result that, when combined, provides the outcome of the rule. Additionally this section describes how rule authors should write test cases, and the mechanism of combining their outcomes.
+Each rule contains one or more testcases written in plain language. The testcases provide the following:
+*Testcase mode identifying whether testcase steps are automated, semi-automated or require manual testing.
+*Testcase steps which clearly outline the steps necessary to reach the result.
+*Test case result of whether testcase passed or failed.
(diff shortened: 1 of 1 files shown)


moekraft:
Question: How do I nest lists? I tried the recommended Markdown format but list did not nest. See http://commonmark.org/help/tutorial/10-nestedLists.html. Does bikeshed support nested lists?

maryjom:
Added to the [22 Feb survey](https://www.w3.org/2002/09/wbs/93339/ACTTF22Feb2017/).

act-framework.bs
```diff
@@ -131,7 +131,15 @@ Selector syntax depends on the input type. When the input data is an HTML docume
...
 
-Editor note: This section describes how rules are broken down into one or more test cases. Each test case gives some result that, when combined, provides the outcome of the rule. Additionally this section describes how rule authors should write test cases, and the mechanism of combining their outcomes.
+Each rule contains one or more testcases written in plain language. Each testcase MUST provide the following:
+
+* Testcase mode identifying whether testcase steps are automated, semi-automated or require manual testing.
+* Testcase steps which clearly outline the steps necessary to reach the result.
+* Test case result indicating whether testcase passed or failed.
+  * Description of error causing any failing result.
+
+Test case results are combined with other test case results to provide a cumulative outcome, pass or fail, of the rule.
```
WilcoFiers: I suggest we update this to the following:

"Test case results are combined with other test case results to provide a cumulative outcome, pass or fail, of the rule. When a rule has multiple test cases, the results of all test cases are combined to give a single for each selected item. The results are cumulative, meaning that as long as one test case passes, the rule has passed.

Example:
A rule for HTML img elements has two test cases, one that checks if there is a text alternative, and another that checks if the image is marked as decorative. If either one of these passed, the rule has passed. Only if both fail, does that element fail the rule."

act-framework.bs
```diff
@@ -131,7 +131,15 @@ Selector syntax depends on the input type. When the input data is an HTML docume
 Test Cases {#test-proc-cases}
 ---------------------------------------
 
-Editor note: This section describes how rules are broken down into one or more test cases. Each test case gives some result that, when combined, provides the outcome of the rule. Additionally this section describes how rule authors should write test cases, and the mechanism of combining their outcomes.
+Each rule contains one or more testcases written in plain language. Each testcase MUST provide the following:
+
+* Testcase mode identifying whether testcase steps are automated, semi-automated or require manual testing.
+* Testcase steps which clearly outline the steps necessary to reach the result.
+* Test case result indicating whether testcase passed or failed.
+  * Description of error causing any failing result.
```
WilcoFiers: Should be "of the error".

act-framework.bs
```diff
@@ -131,7 +131,18 @@ Selector syntax depends on the input type. When the input data is an HTML docume
 Test Cases {#test-proc-cases}
 ---------------------------------------
 
-Editor note: This section describes how rules are broken down into one or more test cases. Each test case gives some result that, when combined, provides the outcome of the rule. Additionally this section describes how rule authors should write test cases, and the mechanism of combining their outcomes.
+Each rule contains one or more testcases written in plain language. Each testcase MUST provide the following:
```
moekraft: From today's meeting we are going with test cases. I will check the document, update and merge the PR. Thanks, Moe

act-framework.bs
```diff
@@ -131,7 +131,18 @@ Selector syntax depends on the input type. When the input data is an HTML docume
...
+
+* Testcase mode identifying whether testcase steps are automated, semi-automated or require manual testing.
+* Testcase steps which clearly outline the steps necessary to reach the result.
+* Test case result indicating whether testcase passed or failed.
+  * Description of the error causing any failing result.
+
```
moekraft: From today's meeting, the decision was that did not make sense under test cases but is covered elsewhere.

PR #57, opened 2017-02-22
Implementation validation

WilcoFiers:


act-framework.bs
```diff
@@ -212,7 +212,9 @@ Editor note: This section will describe how a rule author should document scenar
 Implementation Validation {#quality-implement}
 ----------------------------------------------
 
-Editor note: This section describes the requirements of tests that have to be created for a rule. Rules are abstract, high level descriptions. To ensure the implementation of rules is done correctly, validation tests have to be provided along with each rule.
+Implementation tests are tests for accessibility test tools. These tests enable the developers and users of ATTs to validate the implementation of an ACT Rule. Each rule must have implementation tests for the selector, as well as for each test case in the rule.
+
+An implementation test consists of two parts. It has a snippet of input, such as an HTML code snippet. This piece is put through the rule. The result of this is then compared to the expected result. The expected result is a list of pointers, along with the outcome (pass, fail, inapplicable) the rule should have been found for the element.
```
rdeltour: Proposed wordsmithing:

> An implementation test consists of two parts: a piece of input data and an expected result. When applying the test, the piece of input data –for instance an HTML code snippet– is evaluated by following the rule's test procedure. The result is then compared to the expected result of the test. The expected consists of a list of pointers and the expected outcome (pass, fail, inapplicable) of the evaluation.

PR #58, opened 2017-02-22
Added document abstract

WilcoFiers:


act-framework.bs
```diff
@@ -8,7 +8,7 @@ Status: w3c/ED
 Group: Accessibility Conformance Testing Task Force
 Editor: Wilco Fiers, Deque Systems
 Editor: Maureen Kraft, IBM Corp.
-Abstract: To do
+Abstract: The Accessibility Conformance Testing Framework 1.0 (ACT Framework 1.0) is a specification designed to harmonize how accessibility rule are described for automated test tools, and how test procedures are written for quality assurance testing. By making these rules interchangable, organizations are better document and share their method of testing. An interchangable format is the first step in developing a harmonized approach for accessibility conformance testing.
```
rdeltour: editorial: s/how accessibility rule are/how accessibility rules are/

rdeltour: can't parse:
> By making these rules interchangable, organizations are better document

do you mean "have better document" or "are better documented" or something else?

rdeltour: also, not sure that "interchangeable" is the right concept here. how about something like:

> By adopting a standardized rule description format, organizations are better...

and later:

> A common open format is the first step...

(please disregard if "interchangeable" is indeed a better term, I'm not a native speaker ;-)

Issue #83, opened 2017-05-08
Accuracy benchmarking repeats false positives (editorial)


moekraft:
Section 7.2 Accuracy Benchmarking reads [https://w3c.github.io/wcag-act/act-rules-format.html#quality-accuracy](url) has statement: "The accuracy of a rule is the average between the false positives and false positives, which are in tern calculated as follows:"

Should be "between the false positives and false negatives"  Also note spelling error, "in tern" should be "in turn".

moekraft:
Created PR #84

PR #84, opened 2017-05-12
Fix double false positive statements

moekraft:
Changed second false positives to false negatives. Also, believe we want percentage (singular) vs. percentages (plural).


moekraft:
Addressing Issue #83

Issue #86, opened 2017-05-15
Update Abstract so as not to focus only on Automated Test Tools


moekraft:
Per issue #82 addressing feedback from Raph de Rooij the Abstract has too much of a focus on Automated Test Tools. Rewrite Abstract to call out manual and semi-automated tests.

WilcoFiers:
Resolved in PR #90.

PR #90, opened 2017-05-16
Update act-rules-format.bs

nitedog:
Suggested edits to address issue #86


WilcoFiers:
I like it, looks good to me @nitedog

moekraft:
I agree. This looks good. Although it is probably more grammatically correct, "automatically" and "semi-automatically" don't necessarily flow off the tongue.

Issue #88, opened 2017-05-15
Update outcome


moekraft:
Update the terminology for cannot tell. Shadi suggested "undetermined" Note: this is also applicable to 6.2. 
<shadi> "The result of the test is undetermined. This may be because human intervention or further testing is needed."

dylanb:
I prefer "incomplete" because for semi-automated tests, there may have been some filtering/selection/information extraction performed by the automation that then has to be completed by a human.

Also: this is what we currently call it in axe-core for those reasons

nitedog:
The suggestion was to reword the definition rather than the term itself. Suggestion from the [minutes](https://www.w3.org/2017/05/15-wcag-act-minutes.html):

> "The result of the test is undetermined. This may be because human intervention or further testing is needed."

I think "incomplete" is ambiguous (is the result or the test incomplete?), and it is only one of the possible reasons for ending up with the result "cannot tell".

Note: We also agreed that "not applicable" and "not tested" will be added to the list, but this is not clear from the minutes.

WilcoFiers:
I prefer `undetermined` over `incomplete` when it comes to the state of a result. It seems to me that a **test** is **incomplete**, as long as some of the **outcomes** are **undetermined**. A test, in this use of the word, is the proces of going through the procedure of one or more ACT rules. Each rule returns one or more results, which may be `undetermined` because the procedure could not be completed without further input (such as human evaluation).

cpandhi:
Will work with Moe to update  as per our discussion on June 2.

moekraft:
Addressed via Pull Request #92

PR #92, opened 2017-06-02
Outcomes undetermined

moekraft:
Updated outcome of "cannot tell" to "undetermined"

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -240,7 +240,7 @@ The procedure of a rule MUST always return one of the following outcomes:
 
 - **Passed**: The Test Target passed the test
 - **Failed**: The Test Target failed the test
-- **Undetermined**: The result of the test is undetermined. Human intervention or further testing is needed.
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -240,7 +240,7 @@ The procedure of a rule MUST always return one of the following outcomes:
 
 - **Passed**: The Test Target passed the test
 - **Failed**: The Test Target failed the test
-- **Cannot tell**: It is unclear whether or not the Test Target passes of rails the test
+- **Undetermined**: The result of the test is undetermined. Human intervention or further testing is needed.
```
maryjom: In the definition, you should use a term other than the term you are trying to define, not "undetermined".

cpandhi: Will update the statement to read more clearly, The result of the test is undetermined because Human intervention or further testing is needed.

moekraft:
This addresses issue #88

Issue #93, opened 2017-06-12
Versioning of rules


moekraft:
Per meeting on 6/12:

- Remove versioning section and make it a section on Change Log
- Also, Don't necessarily need to use X.Y.Z versioning instead a date is sufficient

Summary of minutes on this topic;

<rdeltour> wilco: if you have a change log, you might not need a version number
[10:28] <rdeltour> shadi: I think we'll need both
[10:28] <rdeltour> ... some people with say version X of the test rules, but some other want to know about individual rules
[10:28] <Sujasree> <sujasree> +1 to wilco... it is sufficient to have change log and do not require version number on individual rule file
[10:28] <rdeltour> ... we might want to differentiate major and minor changes
[10:29] <rdeltour> ... how much of a change can you still bear to be the same rule
[10:29] <rdeltour> ... there might be a variety of changes, technology emerges, AT change, etc. all this can influence a rule
[10:29] <rdeltour> ... and bugs, typos, etc
[10:30] <rdeltour> wilco: so you're saying we should still track versions of individual rules or a changelog would be sufficient?
[10:31] <rdeltour> shadi: I think a version w/b useful, but we also need a separate process for more major changes
[10:31] <rdeltour> ... at w3c we use dates as version
[10:31] <rdeltour> ... I don't mind whether it's a version number, or a date, etc.
[10:32] <rdeltour> ... creating a changelog is quite simple
[10:32] <rdeltour> wilco: what do you tink of rolling the versions in the changelog itself?
[10:32] <rdeltour> shadi: it can work. I think it's enough if there's a date. we can also auto-generate a diff
[10:33] == Charu [~Charu@public.cloak] has joined #wcag-act
[10:33] <rdeltour> ... we can simplify the authoring with an automated diff service

PR #94, opened 2017-06-16
Merge change log with version history.

WilcoFiers:


moekraft:
From today's call, 6/19: Make this less prescriptive and add examples. "Significant changes, for example those that change the scope of a rule, may require the creation of a new rule and deprecation of the current rule." Give 2 examples major and minor changes.

moekraft:
This pull request addresses Issue #93

moekraft:
@WilcoFiers Did we agree to merge this PR in light of the 2 new issues from today's discussion?

act-rules-format.bs
```diff
@@ -304,22 +304,14 @@ See [[#appendix-data-example]] on how this could be expressed using JSON-LD and
...
 
-: Z / Patch updates:
-:: The patch version number must be increased if the change does not affect the result of a rule. This includes editorial changes, new issues on the issues list, an updated description, etc.
+Each new release of an ACT Rule MUST be identifiable with either a date or a version number. Additionally, a reference to the previous version of that rule MUST be available. It is recommended that for huge changes, a new rule is created and the old rule is deprecated. 
 
-### Change Log ### {#quality-updates-changes}
+An example of when a new rule should be created would be when going from a rule that tests the use of a blink element, and changes it to a rule that looks for animated style changes. This potentially adds lots of new issues that were previously out of scope. But for that same rule, adding a step to check if the blink element is positioned off screen, should be done by updating the existing rule.
```
annethyme: consider removing "and changes it"

act-rules-format.bs
```diff
@@ -304,22 +304,14 @@ See [[#appendix-data-example]] on how this could be expressed using JSON-LD and
...
-:: The major version number must be increased if the change can lead to new failure results.
+### Change Log ### {#quality-updates-changes}
 
-: Y / Minor updates:
-:: The minor version number must be increased if the test logic has been updated, which could lead to a a different result.
+It is important to keep track of what has changed in ACT rules, so that users of them can work out if changes in accessibility results are due to changes in the rules used in testing, rather than changes in the content itself. All changes to an ACT Rule that can change the outcome of a test MUST be recorded in a change log. The change log can either be part of the rule document itself, or be referenced from it.
 
-: Z / Patch updates:
-:: The patch version number must be increased if the change does not affect the result of a rule. This includes editorial changes, new issues on the issues list, an updated description, etc.
+Each new release of an ACT Rule MUST be identifiable with either a date or a version number. Additionally, a reference to the previous version of that rule MUST be available. It is recommended that for huge changes, a new rule is created and the old rule is deprecated.
```
annethyme: consider using another word instead of "huge"

maryjom: Agree with the comment on "huge". Suggest maybe using "extensive".  Why would you deprecate a rule for large changes rather than update the major version number?

act-rules-format.bs
maryjom: The problem with this example is that you wouldn't want to deprecate the original rule, as it would still be a valid test for the blink element.  You would, however want to create a new rule for the animated element checking - that part makes sense.

2biazdk: Agree that it would make more sense grammatically to have it say:

> (...) going from a rule that (...), to a rule that (...)

act-rules-format.bs
moekraft: I agree with Mary Jo. I like extensive. Also, I have edits on the first paragraph:

"It is important to keep track of changes to the ACT rules so that users of the rules can understand if changes in test results are due to changes in the rules used when performing the tests, rather than changes in the content itself. All changes to an ACT Rule that can change the outcome of a test MUST be recorded in a change log. The change log can either be part of the rule document itself or be referenced from it.

Issue #87, opened 2017-05-15
Update Scope with mention of Mapping rules to Success Criteria


moekraft:
[10:19] <ChrisLoiselle> Wilco : Do we want to include partial test results text into introduction as well? Combining different test results...abstract...
[10:20] <ChrisLoiselle> Shadi : A rule doesn't necessarily test one success criteria, could include many within one test
[10:21] <ChrisLoiselle> Wilco: Rule aggregation already covers that.
[10:22] <ChrisLoiselle> Shadi: this should be a heading of its own. Relationship between test rules and WCAG SC's. Or broaden criteria. Mapping between test rules and SCs. Needs own heading.
[10:22] <ChrisLoiselle> Wilco: Mentions we can talk to this in section 3.3.
[10:23] <ChrisLoiselle> Shadi: either in introduction or scope would be better, more general and not inside test rule
[10:23] <ChrisLoiselle> Shadi: use "Mapping" vs. "relate".
[10:24] <ChrisLoiselle> MoeKraft: will create ticket on creating section within scope on mapping between rules and success criteria
[10:24] <ChrisLoiselle> MoeKraft: will assign that to Wilco

moekraft:
Closed by proxy via Issue #96

PR #96, opened 2017-07-07
Expand accessibility requirements section

WilcoFiers:
This PR addresses #87

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -59,11 +59,11 @@ Each ACT Rule MUST have a description that is in plain language and provides a b
 Accessibility Requirements {#structure-accessibility-requirements}
 -------------------------------------------------------------------
 
-This section should explain the accessibility requirement the rule maps to, (for example, WCAG 2.0 success criterion 1.1.1). An ACT Rule DOES NOT have to test the entire accessibility requirement. The Rule COULD be used to test multiple accessibility requirements, when such requirements overlap each other.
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -59,7 +59,9 @@ Each ACT Rule MUST have a description that is in plain language and provides a b
 Accessibility Requirements {#structure-accessibility-requirements}
 -------------------------------------------------------------------
 
-Explain the accessibility requirement being tested such as the WCAG 2 Success criterion and / or the technique the rule maps to; For example WCAG 2.0 Technique H67.
+Explain the accessibility requirement being tested such as the WCAG 2 Success criterion and / or the technique the rule maps to; For example WCAG 2.0 Technique H67. An ACT Rule DOES NOT have to test the entire accessibility requirement. The Rule COULD be used to test multiple accessibility requirements, when such requirements overlap each other.
+
```
kasperisager: `WCAG 2` => `WCAG 2.0`, no?

WilcoFiers: Could also be WCAG 2.1 criteria once it is released. It doesn't matter much.

moekraft:
This section explains the accessibility requirements to which the rule maps, (for example, WCAG 2.0 success criterion 1.1.1). An ACT Rule DOES NOT have to test the entire accessibility requirement. The Rule COULD be used to test multiple accessibility requirements.

Outcomes from an ACT Rule MUST be consistent within a rule, e.g. if the rule results in a failure, the failure MUST be for the defined accessibility requirement. This means that the rule maps to the accessibility requirement, as opposed to it merely being related to the requirement, thematically or otherwise.

Note: For WCAG 2.0, the Success Criteria are the accessibility requirements. Often organizations have accessibility requirements in addition to the WCAG success criteria. These too can be tested using ACT Rules.

PR #143, opened 2017-11-23
Add note on unique identifiers

### LINKED ISSUE #133: Define "Unique identifier"
A _unique identifier_ can be many things, and the spec currently does not define what it is. It should clarify for instance whether it has to be universally unique, or just within the context of a rule set.

WilcoFiers:
This closes #133

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -49,6 +49,7 @@ Each ACT Rule MUST provide the following items written in plain language:
 * [Test Subject Type](#input-types)
 * [Test Procedure](#test-proc)
 
+<p class="note">Note: The unique identifier can be any type of text or number, such as a URI or a number in the database.</p>
 
 Rule Description {#structure-description}
 ----------------------------------------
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -49,6 +49,7 @@ Each ACT Rule MUST provide the following items written in plain language:
 * [Test Subject Type](#input-types)
 * [Test Procedure](#test-proc)
 
+<p class="note">Note: The unique identifier can be any type of text or number, such as a URI or a number in the database.</p>
```
rdeltour: > a URI

I would say URL. I know it is sometimes controversial, but the most recent an unifying spec is the WhatWG's URL standard, which does deprecate the URI term in favor of URL, even for identifiers.

> a number in the database

it reads very slightly unclear: which database?

Suggested rewording:

> The unique identifier can be any text string, such as for instance a URL or a database identifier.

moekraft: From 11/30/17 meeting: "The unique identifier can be any unique text value, such as plain text, URL or a database identifier."

PR #168, opened 2018-02-01
Reword the abstract

rdeltour:
- remove the use of "requirements" (requirements for whom?)
- replace "organizations" by "any party involved in accessibilty testing"
  (since this format may not only be used by organizations, but also
   by individuals)
- replace "better" by "in a robust and undertandable manner" to be more
  explicit

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -8,7 +8,7 @@ Status: ED
 Group: act-framework
 Editor: Wilco Fiers, Deque Systems
 Editor: Maureen Kraft, IBM Corp.
-Abstract: The Accessibility Conformance Testing (ACT) Rules Format 1.0 specifies requirements for writing accessibility test rules. This includes test rules that are carried out fully-automatically, semi-automatically, and manually. By defining this common format, organizations are better able to document and share their testing procedures. This enables transparency and harmonization of testing methods, including methods implemented by accessibility test tools.
+Abstract: This specification defines a format for writing accessibility test rules. These rules can be carried out fully-automatically, semi-automatically, and manually. This common format allows any party involved in accessibility testing to document and share their testing procedures in a robust and understandable manner. This enables transparency and harmonization of testing methods, including methods implemented by accessibility test tools.
 Markup Shorthands: markdown yes
 </pre>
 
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -8,7 +8,7 @@ Status: ED
 Group: act-framework
 Editor: Wilco Fiers, Deque Systems
 Editor: Maureen Kraft, IBM Corp.
-Abstract: The Accessibility Conformance Testing (ACT) Rules Format 1.0 specifies requirements for writing accessibility test rules. This includes test rules that are carried out fully-automatically, semi-automatically, and manually. By defining this common format, organizations are better able to document and share their testing procedures. This enables transparency and harmonization of testing methods, including methods implemented by accessibility test tools.
+Abstract: This specification defines a format for writing accessibility test rules. These rules can be carried out fully-automatically, semi-automatically, and manually. This common format allows any party involved in accessibility testing to document and share their testing procedures in a robust and understandable manner. This enables transparency and harmonization of testing methods, including methods implemented by accessibility test tools.
```
WilcoFiers: I think we should keep the first sentence. I get that it's a bit repetative, but it's good to have the name in it.

rdeltour: Yeah, what I didn’t like about the first sentence was the "specifies requirements" bit, which feels obvious for a spec ;-) and the fact that the essence of the spec is to define a format so I wanted to make that appear more clearly.

If we keep the name of the spec, it reads "The Accessibility Conformance Testing (ACT) Rules Format 1.0 defines a format for writing accessibility test rules" which is a bit repetitive but OK. Some other W3C specs don’t have the name in the abstract (like for instance HTML), some do (like WCAG).

So your call, boss :-)

WilcoFiers: Yeah, Shadi suggested we keep it so it can be coppied as a stand-alone text. Can you put in a quick update so I can merge? I want to get a CfC out before the weekend.

rdeltour: Sure can do. So I’ll keep the spec but keep also the reword in the second part of the sentence? like so:

> The Accessibility Conformance Testing (ACT) Rules Format 1.0 defines a format for writing accessibility test rules

Can you confirm?

PR #171, opened 2018-02-01
Reword the Accessibility Support section

rdeltour:
- remove "good and bad", which is mostly subjective
- reword "This is known" ("this" referred to what?)
- remove use of "may" and "should" in explanatory prose
- normalize the normative statements on ACT Rules

Also, I’m wondering if "significant accessibility support concerns" can be made more explicit?

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -76,7 +76,7 @@ List any limitations, assumptions or any exceptions for the test, the test envir
 Accessibility Support {#acc-support}
 ===========================
 
-ACT Rules are designed to test accessible uses of a web technology. However, not every part of a web technology is implemented in all assistive technologies a website may need to support. The concept of [accessibility supported](https://www.w3.org/TR/WCAG20/#accessibility-supporteddef) use of a Web technology is defined in [[WCAG20]]. Because of this, ACT Rules are not necessarily applicable in all test scenarios. For instance, a rule related to ARIA could return false-negative results on a site that is supposed to work for assistive technologies that have no ARIA support.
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -76,11 +76,11 @@ List any limitations, assumptions or any exceptions for the test, the test envir
 Accessibility Support {#acc-support}
 ===========================
 
-ACT Rules are designed to test good and bad uses of a web technology. Not every part of a web technology is implemented in all assistive technologies a website may need to support. This is known as [Accessibility Supported in WCAG 2.0](https://www.w3.org/TR/WCAG20/#accessibility-supporteddef). Because of this, some rules should not be used in a particular test scenario. For instance, rules related to ARIA should not be used on a site that is supposed to work for assistive technologies that have no ARIA support.
+ACT Rules are designed to test accessible uses of a web technology. However, not every part of a web technology is implemented in all assistive technologies a website may need to support. The concept of [accessibility supported](https://www.w3.org/TR/WCAG20/#accessibility-supporteddef) use of a Web technology is defined in [[WCAG20]]. Because of this, ACT Rules are not necessarily applicable in all test scenarios. For instance, a rule related to ARIA could return false-negative results on a site that is supposed to work for assistive technologies that have no ARIA support.
```
WilcoFiers: I think this last sentence should be:

> For instance, a web page that has to work in assistive technologies that have no WAI-ARIA support, shouldn't be tested with an ACT Rule that relies on WAI-ARIA support, since this could lead to false positive results.

rdeltour: Ok for turning it like this, but we need to remove the "shouldn’t" as it’s an example and not a conformance statement. We could use "wouldn’t" or "can’t", not exactly the same, but maybe just fine enough.

WilcoFiers: Works for me. Please update so I can merge put it in the CFC draft.

PR #190, opened 2018-03-17
Fix occurrences of typo using the contraction of 'it is' instead of the possessive 'its'.

enummela:


WilcoFiers:
Thanks @enummela, much appreciated!

PR #199, opened 2018-04-19
Update Introduction Section

anthonyFer:
1. We need to be consistent with either with the word "products"  or "tools" used in this section.
2. Removed additional words that add cognitive load the sentence.
3. Usually, in QA terms we say "Testing Tools" instead of "Test Tools".

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -15,7 +15,7 @@ Markup Shorthands: markdown yes
 Introduction {#intro}
 =====================
 
-There are currently many tools available which aid their users in testing web content for conformance to the Web Content Accessibility Guidelines (WCAG). As the web develops in both size and complexity, these tools are essential for managing the resources available on the web.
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -15,9 +15,9 @@ Markup Shorthands: markdown yes
 Introduction {#intro}
 =====================
 
-There are currently many products available which aid their users in testing web content for conformance to accessibility standards such as the Web Content Accessibility Guidelines (WCAG) 2.0. As the web develops and grows in both size and complexity, these tools are essential for managing the accessibility of resources available on the web.
+There are currently many tools available which aid their users in testing web content for conformance to the Web Content Accessibility Guidelines (WCAG). As the web develops in both size and complexity, these tools are essential for managing the resources available on the web.
```
WilcoFiers: @anthonyFer The following should not be removed "accessibility standards such as "

PR #200, opened 2018-04-19
Update Accessibility Support Section

anthonyFer:
1. The sentence "to test accessible uses of a web technology" is confusing. 
2. The word "that" was missing?


moekraft:
Reviewed at 5/3 meeting. Anthony will continue to work on the wording. Take a look at definition of Accessibility Supported for inspiration: https://www.w3.org/TR/WCAG20/#accessibility-supporteddef

act-rules-format.bs
```diff
@@ -80,7 +80,7 @@ An ACT Rule MUST list any limitations, assumptions or any exceptions for the tes
...
+ACT Rules are designed to test the accessibility conformance in applications of web technologies. However, not every part of a web technology is implemented in all assistive technologies that a website may need to support. The concept of [accessibility supported](https://www.w3.org/TR/WCAG20/#accessibility-supporteddef) use of a Web technology is defined in [[WCAG20]]. Because of this, ACT Rules are not necessarily applicable in all test scenarios. For instance, a web page that has to work in assistive technologies that have no WAI-ARIA support, wouldn’t be tested with an ACT Rule that relies on WAI-ARIA support, since this could lead to false positive results.
```
kasperisager: The casing of "web technology" is not entirely consistent.

maryjom: It isn't just about the implementation in assistive technology for accessibility support. Per the WCAG definition, it's also about user agent support of the accessibility features in the technology.  For example, use of technically sound WAI-ARIA may not be fully supported in older user agents or older assistive technology. In either case, the accessibility support isn't there. So are rules expected to always have defined the parameters (AT type, version and browser type, version) where there is accessibility support? Or for the user agent support can they simply point to the documented test results (as an example, the ARIA 1.1 support results).

WilcoFiers: This is still saying that rules test conformance of web technologies. That's not correct. Rules test conformance of applications built with web technologies.

PR #211, opened 2018-05-09
Update rule grouping section

kasperisager:
@annethyme

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -66,7 +66,9 @@ Accessibility Requirements {#structure-accessibility-requirements}
 
 An ACT Rule MUST identify the accessibility requirements to which the rule maps. For example, WCAG 2.0 Success Criterion 1.1.1. An ACT Rule is a complete or partial test for one or more accessibility requirements.
 
-Outcomes from an ACT Rule MUST be consistent with the accessibility requirement, e.g. a rule only returns the outcome Fail when the content fails the accessibility requirement. This means that the rule maps to the accessibility requirement, as opposed to it merely being related to the requirement, thematically or otherwise.
+Outcomes from an ACT Rule that is not part of an [aggregation group](#grouping) MUST be consistent with the accessibility requirement, e.g. a rule only returns the outcome Fail when the content fails the accessibility requirement. This means that the rule maps to the accessibility requirement, as opposed to it merely being related to the requirement, thematically or otherwise.
+
+For ACT Rules that are part of aggregation groups, the aggregated outcome of the aggregation group MUST be consistent with the accessibility requirement. For this to be possible, ACT Rules in an aggregation group MUST share at least one common accessibility requirement.
 
 The actual definition of specific accessibility requirements is beyond the scope of ACT Rules and of this document. For WCAG, Success Criteria are considered to be accessibility requirements. Some organizations have additional accessibility requirements, such as specific implementation techniques to meet WCAG Success Criteria.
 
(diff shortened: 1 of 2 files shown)


kasperisager:
Should we add `act-rules-format.html` to `.gitignore` by the way?

act-rules-format.bs
```diff
@@ -66,7 +66,9 @@ Accessibility Requirements {#structure-accessibility-requirements}
 
 An ACT Rule MUST identify the accessibility requirements to which the rule maps. For example, WCAG 2.0 Success Criterion 1.1.1. An ACT Rule is a complete or partial test for one or more accessibility requirements.
 
-Outcomes from an ACT Rule MUST be consistent with the accessibility requirement, e.g. a rule only returns the outcome Fail when the content fails the accessibility requirement. This means that the rule maps to the accessibility requirement, as opposed to it merely being related to the requirement, thematically or otherwise.
+Outcomes from an ACT Rule that is not part of an [aggregation group](#grouping) MUST be consistent with the accessibility requirement, e.g. a rule only returns the outcome Fail when the content fails the accessibility requirement. This means that the rule maps to the accessibility requirement, as opposed to it merely being related to the requirement, thematically or otherwise.
```
WilcoFiers: "rule only returns the outcome Fail when the content fails the accessibility requirement" That's not correct. Rules can return fail, even if the requirement is passed. If only one rule in a group has to pass for the AR to pass, than any number of rules can fail without the AR failing.

act-rules-format.bs
```diff
@@ -66,7 +66,9 @@ Accessibility Requirements {#structure-accessibility-requirements}
...
+
+For ACT Rules that are part of aggregation groups, the aggregated outcome of the aggregation group MUST be consistent with the accessibility requirement. For this to be possible, ACT Rules in an aggregation group MUST share at least one common accessibility requirement.
```
WilcoFiers: This seems to be repeating what is in the previous paragraph. I don't think this should say "aggregated outcome". It's just "the outcome of the aggregation group" is sufficient.

Question: Should the aggregation groups themselves be associated with accessibility requirements, rather than the rules within them?

act-rules-format.bs
```diff
@@ -170,22 +172,48 @@ When all expectations are true for a *test target* it `passed` the rule. If one
...
 
-Rule Grouping {#grouping}
+Aggregation Groups {#grouping}
 ===================
 
-In accessibility testing, there are often multiple ways to solve the same problem. For instance, header cells in HTML tables can be indicated through the `scope` attribute, by using the `headers` attribute, or by using ARIA labels. All of these separate techniques could be described in one big rule. But this creates a large, and often difficult to maintain rule. To ensure rules are kept small and atomic, they SHOULD be put into a rule-group instead.
+In accessibility testing, there are often different ways to make the same type of content accessible. These different techniques could be described in one big rule. This however creates large, and often difficult to maintain rules.
+
+Rules are kept small and atomic. If there are multiple ways of passing the same accessibility requirement for the same content, these SHOULD be written as separate rules and be put into an aggregation group instead. A rule MAY be part of multiple aggregation groups.
```
WilcoFiers: Rules **should be** kept small and atomic.

act-rules-format.bs
```diff
@@ -170,22 +172,48 @@ When all expectations are true for a *test target* it `passed` the rule. If one
...
-Note that rules in a group MAY have different applicability. Because of this, not every element applicable within the group is tested by every rule. Rules MAY also be disabled during a test run due to accessibility support concerns. See [Accessibility support](#acc-support) for details.
+<div class="example">
+  <p>Aggregation group: Header cells in HTML tables (WCAG 2 Success Criterion 1.3.1)</p>
+  <p>Examples of rules in aggregation group:</p>
+  <ul>
+    <li>Header indicated through the `scope` attribute</li>
+    <li>Header indicated by using the `headers` attribute</li>
+    <li>Header indicated by using ARIA labels</li>
+  </ul>
+  <p>Aggregation logic for group: for a given test target to pass the group, any of the rules within the group must output a passed result for the test target.</p>
```
WilcoFiers: I would suggest "at leats one" instead of "any".

act-rules-format.bs
annethyme: This section is about "ACT Rule that is not part of an [aggregation group](#grouping)". Next section is about "ACT Rules that are part of aggregation groups".

WilcoFiers:
Closing this PR, as it is superseded by #217.

PR #212, opened 2018-05-17
Improve the example of 'Limitations, Assumptions, or Exceptions'

### LINKED ISSUE #136: Example of section 3.4 “Limitations, Assumptions, or Exceptions”
I find the example given in section 3.4 not very obvious:

> For example, A rule for Success Criterion 1.4.1: Use of Color has to make an assumption that CSS is used to make a link visually evident - typically by using CSS background, border, color, font, or text-decoration properties.

The path from SC 1.4.1 to requiring CSS to make a link visually evident isn't obvious for one not versed in WCAG/HTML. Also, a rule could be partially based on SC 1.4.1 and not require parsing CSS at all.

rdeltour:
Change the example of the "Limitations" section to make it more explicit.

Fixes #136.

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -74,7 +74,7 @@ The actual definition of specific accessibility requirements is beyond the scope
 Limitations, Assumptions or Exceptions {#structure-limitations-assumptions-exceptions}
 -------------------------------------------------------------------
 
-An ACT Rule MUST list any limitations, assumptions or any exceptions for the test, the test environment, technologies being used or the subject being tested. For example, a rule for Success Criterion 1.4.1: Use of Color has to make an assumption that CSS is used to make a link visually evident - typically by using CSS background, border, color, font, or text-decoration properties.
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -74,7 +74,7 @@ The actual definition of specific accessibility requirements is beyond the scope
 Limitations, Assumptions or Exceptions {#structure-limitations-assumptions-exceptions}
 -------------------------------------------------------------------
 
-An ACT Rule MUST list any limitations, assumptions or any exceptions for the test, the test environment, technologies being used or the subject being tested. For example, a rule for Success Criterion 1.4.1: Use of Color has to make an assumption that CSS is used to make a link visually evident - typically by using CSS background, border, color, font, or text-decoration properties.
+An ACT Rule MUST list any limitations, assumptions or any exceptions for the test, the test environment, technologies being used or the subject being tested. For example, a rule that would partially test WCAG 2.0 Success Criterion 1.4.3 (minimum contrast) based on the inspection of CSS properties could state that it is only applicable to HTML text content stylable with CSS and does not support images of text.
```
annethyme: WCAG 2.0 --> WCAG 2

annethyme: If we write the whole name of the success criterion here, we probably need to do it consistently throughout the document. 
Also, the full name of 1.4.3 is "Contrast (Minimum)".

Issue #233, opened 2018-07-24
8.1. Common Aspects - Question


annika-FTB:
How will the ACT rules format accommodate rules for the new Success Criteria of WCAG 2.1 (Orientation, Gestures, etc.)? Is there a common way to describe those?

WilcoFiers:
Suggestion to maintain a list of common aspects separate from the spec and only provide examples in the spec itself.

annethyme:
Created https://github.com/w3c/wcag-act/pull/242 as a suggested solution.

WilcoFiers:
@annika-FTB The answer to the question is yes, ACT Rules are not limited to WCAG 2.0. Auto-WCAG is currently working on a few WCAG 2.1 rules. The first one of these is already published here:

https://auto-wcag.github.io/auto-wcag/rules/SC1-3-5-autocomplete-valid.html

maryjom:
Reopened this issue, as we haven't yet surveyed the suggested updates and are doing that this week.

moekraft:
Merged PR #242

PR #242, opened 2018-08-16
Update "8. Aspects under Test"

annethyme:
Remove "8.1 Common Aspects" to move it to a working group note. Insert link and examples in "8. Aspects under Test" instead.
In answer to https://github.com/w3c/wcag-act/issues/233.

***
<a href="https://pr-preview.s3.amazonaws.com/annethyme/wcag-act/pull/242.html" title="Last updated on Sep 13, 2018, 1:39 PM GMT (566ab45)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/242/54fc1ba...annethyme:566ab45.html" title="Last updated on Sep 13, 2018, 1:39 PM GMT (566ab45)">Diff</a>

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -113,29 +113,29 @@ A list of common aspects for reference in ACT rules can be found in [Common Aspe
 
 <div class=example>
   <p>Test aspects for a rule that checks if images have an accessible name:</p>
-  <li>
-    <ul>DOM Tree</ul>
-    <ul>CSS Styling</ul>
-  </li>
+  <ul>
+    <li>DOM Tree</li>
+    <li>CSS Styling</li>
+  </ul>
 </div>
 
 <div class=example>
   <p>Test aspects for a rule that checks if a transcript is available for videos:</p>
-    <li>
-    <ul>DOM Tree</ul
-    <ul>CSS Styling</ul
-    <ul>Audio output</ul
-    <ul>Visual output</ul
-  </li>
+  <ul>
+    <li>DOM Tree</li>
+    <li>CSS Styling</li>
+    <li>Audio output</li>
+    <li>Visual output</li>
+  </ul>
 </div>
 
 <div class=example>
   <p>Test aspects for a rule that checks for use of (language specific) generic link texts like "click here" and "more":</p>
-    <li>
-    <ul>DOM Tree</ul
-    <ul>CSS Styling</ul
-    <ul>Language</ul
-  </li>
+    <ul>
+    <li>DOM Tree</li>
+    <li>CSS Styling</li>
+    <li>Language</li>
+  </ul>
 </div>
 
 Test Definition (Atomic rules only) {#test-def}
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -109,37 +109,25 @@ An aspect is a distinct part of the [test subject](#output-test-subject) or its 
 
 An atomic rule MUST include a description of all the aspects under test by the rule. Each aspect MUST be discrete with no overlap between the aspects. Some aspects are already well defined within the context of web content, such as HTTP messages, DOM tree, and CSS styling [[CSS2]], and do not warrant a detailed description. Other aspects may not be well defined or even specific to web content. In these cases, an ACT Rule SHOULD include either a detailed description of the aspect in question or a reference to that description.
 
+A list of common aspects for reference in ACT rules can be found in [Common Aspects under Test](#).
 
-Common Aspects {#input-aspects-common}
---------------------------------------
+<div class=example>
+  <p>Test aspects for a rule that checks if images have an accessible name:</p>
+  <li>
+    <ul>DOM Tree</ul>
+    <ul>CSS Styling</ul>
+  </li>
+</div>
 
-### HTTP Messages ### {#input-aspects-http}
-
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -109,38 +109,34 @@ An aspect is a distinct part of the [test subject](#output-test-subject) or its
 
 An atomic rule MUST include a description of all the aspects under test by the rule. Each aspect MUST be discrete with no overlap between the aspects. Some aspects are already well defined within the context of web content, such as HTTP messages, DOM tree, and CSS styling [[CSS2]], and do not warrant a detailed description. Other aspects may not be well defined or even specific to web content. In these cases, an ACT Rule SHOULD include either a detailed description of the aspect in question or a reference to that description.
 
+A list of common aspects for reference in ACT rules can be found in [Common Aspects under Test](#).
```
WilcoFiers: @nitedog can you create a page for us to link to somewhere on w3.org?

nitedog: how do you feel about a WG Note instead?

moekraft: I think that works. @maryjom any thoughts?

maryjom: @nitedog Can you explain what the difference would be between a WG Note, an Understanding doc, and simply putting it in a non-normative Appendix.  Just trying to minimize the number of documents we need to maintain and understand the right place to develop/put such content.

nitedog: A WG Note is non-normative. This is fairly easy to write, publish, and update,, and is more formal than "just any page on the W3C website". At the same time, it is not part of the primary ACT Rules Format specification, which is a normative W3C Recommendation and significantly harder to update (eg. when we identify new aspects).

Actually, I foresee three or four WG Notes supporting the primary ACT Rules Format specification:
- Collection of "Aspects" (eg. HTTP etc. taken from the current spec)
- Generic review process (which we already have https://w3c.github.io/wcag-act-rules/review-process.html)
- Suggested notation for output format (which we also have at https://github.com/w3c/earl)
- (possibly) suggested template for rules (also available https://auto-wcag.github.io/auto-wcag/pages/design/rule-template.html)

All these WG Notes would be referenced in a non-normative way (eg. "you MAY use the following" type of requirement). We already have most of the content but it is rather scattered. This would pull all the content into one "package" with coherent cross-referencing and linking.

Could you please add this for discussion at one of the next calls?

moekraft:
Added link to new WG Note https://w3c.github.io/wcag-act/NOTE-act-rules-common-aspects.html

moekraft:
Diff View of pull request: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwcag-act%2Fact-rules-format.html&doc2=https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/w3c/wcag-act/7723b92d89d69478349526538ba0026c0d8d0fa9/act-rules-format.bs

nitedog:
I was not able to add this as a PR on this PR, so there is the suggested rewording:

An aspect is a distinct part of the [Test Subject](#output-test-subject) or its underlying implementation. For example, rendering a particular piece of content to an end user can involve multiple different technologies, some or all of which can be of interest to an ACT Rule. Some rules need to operate directly on the [Hypertext Transfer Protocol](https://tools.ietf.org/html/rfc7230) [[http11]] messages exchanged between a server and a client, while others need to operate on the [Document Object Model](https://dom.spec.whatwg.org) [[DOM]] tree exposed by a web browser. Other rules need to operate on several aspects simultaneously, such as both the HTTP messages and the DOM tree.

[=Atomic rules=] MUST list the aspects used in the [Test Definition](#test-def). Each aspect MUST be discrete with no overlap between the aspects.

An atomic rule MUST include a description of all the aspects under test by the rule. Some aspects are already well defined within the context of web content, such as HTTP messages, DOM tree, and CSS styling [[CSS2]], and do not warrant a detailed description. Other aspects may not be well defined or even specific to web content. In these cases, an ACT Rule SHOULD include either a detailed description of the aspect in question or a reference to that description.

A list of common aspects for reference in ACT rules can be found in [Common Aspects under Test](https://w3c.github.io/wcag-act/NOTE-act-rules-common-aspects.html).

Since there is no Test Definition in Composed rules, there SHOULD NOT be an aspects under test list for composed rules.

act-rules-format.bs
```diff
@@ -109,38 +109,34 @@ An aspect is a distinct part of the [test subject](#output-test-subject) or its
...
+    <li>
+    <ul>DOM Tree</ul
+    <ul>CSS Styling</ul
+    <ul>Audio output</ul
+    <ul>Visual output</ul
+  </li>
+</div>
 
+<div class=example>
+  <p>Test aspects for a rule that checks for use of (language specific) generic link texts like "click here" and "more":</p>
```
annethyme: Should this be switched out for an example where there is less controversy about how it should work or if a test for this can actually be created? 

Another possible example that would have the same test aspects: 

"Test aspects for a rule that checks if the language attribute for a web page matches the human language of the text on that page."

tbostic32: I personally like the current example of checking for a transcript since it concedes that defining the test aspects may not always be black and white, and best judgement may be necessary. 

My only worry is that since it is an example in our document it may be seen as a "fully-complete" or "fully-correct" case that needs no further tweaking and will be referred to during implementation later on.

annethyme:
@moekraft, I think I got the styling issues in the examples fixed now.

moekraft:
@annethyme Thank you!

PR #259, opened 2018-09-10
Create archive version

moekraft:


moekraft:
Creating an archive version of common aspects WG note.

PR #273, opened 2018-09-21
Bunch of editorial edits

### LINKED ISSUE #220: Concern about stability of test target pointers
Given [this paragraph of the current spec](https://w3c.github.io/wcag-act/act-rules-format.html#output-test-subject)...

> When a single URL can be used to reference the web page, or other test subject, this URL MUST be used.

...I'm somewhat concerned about the stability of [pointers to test targets](https://w3c.github.io/wcag-act/act-rules-format.html#output-test-target). If we require that a URL be used to identify the test subject, rather than allow tools to embed the test subject in its entirety, we run into the issue that the pointers of a given result set are likely to become invalid, or even incorrect, as the test subject undergoes change. This is especially true for non-expression based pointers.

I'd strongly recommend that the "MUST" be changed to "SHOULD" or even the entire sentence left out. I don't really see a reason as to why we shouldn't simply refer to the [`TestSubject`](https://www.w3.org/TR/EARL/#TestSubject) class defined in EARL.

### LINKED ISSUE #230: 3. ACT Rule Types - Clarify example
The example "Header cells in HTML tables" states: 
"If any one of these rules passes, the table cell can pass the composed rule."

The wording "**can** pass the rule" is unclear.
If there are any additional situations, in which the rule is not passed although all the atomic rules are passed, then this should be stated explicitly. Otherwise the sentence should be changed to: "If any one of these rules passes, the table cell passes the composed rule."

### LINKED ISSUE #232: 4. ACT Rule Structure - Improvement suggestion
Add an item in the rule structure that describes the type of the rule (atomic or composed). In this way the types can be distinguished easily without having the refer to indirect information provided in other items such as the "aspects" or "rules list" items.

### LINKED ISSUE #239: False negatives definition
In the section 15.2. Accuracy Benchmarking, false positive and negative are defined : 

> False positives: This is the percentage of test targets, that were failed by the rule, but were not failed by an accessibility expert.
> 
 > False negatives: This is the percentage of test targets, that were passed by the rule, but were failed by an accessibility expert.

To be consistent with the definition of a false positive, I would say for false negatives : "_not passed_" instead of "_failed_" because a test target can be missed by the accessibility expert which will evaluate the rule to "innaplicable" and not "failed" ?

Proposal :  

> _False negatives: This is the percentage of test targets, that were passed by the rule, but were *not passed* by an accessibility expert._

### LINKED ISSUE #257: Aspects Under Tests - Editorial updates (Shadi)
8.0 Aspects Under Test
An aspect is a distinct part of the test subject or its underlying implementation. For example, rendering a particular piece of content to an end user involves multiple different technologies, some or all of which may be of interest to an ACT Rule. Some rules need to operate directly on the Hypertext Transfer Protocol [http11] messages exchanged between a server and a client, while others need to operate on the Document Object Model [DOM] tree exposed by a web browser.
Atomic rules MUST list the aspects used in the Test Definition. Some rules may need to operate on several aspects simultaneously, such as both the HTTP messages and the DOM tree.
An atomic rule MUST include a description of all the aspects under test by the rule. Some aspects are already well defined in a formal specification within the context of web content, such as HTTP messages, DOM tree, and CSS styling [CSS2]. These do not warrant a detailed description further than a reference to the corresponding section in this specification (see Common Aspects below). For other aspects that are not well defined, an ACT Rule SHOULD include either a detailed description of the aspect in question or a reference to that description.
Each aspect MUST be discrete with no overlap between the aspects.
There SHOULD NOT be an aspects under test list for composed rules.
[Accessibility Conformance Testing.docx](https://github.com/w3c/wcag-act/files/2357188/Accessibility.Conformance.Testing.docx)

### LINKED ISSUE #264: Clarify definition of Aspect under Test needed
During weekly WCAG ACT TF meeting, an issue was raised regarding what is meant by Aspect under test.

### LINKED ISSUE #266: Doc references "#output-aggregation" uses different link text each time used
I've found a few instances of an anchor link used "#output-aggregation" where the link text is different each time, and I think it should be consistent.  Occurrences found are:

- Most places, says 'Rule Aggregation'
- In Section 14.3 Outcome (see first green highlighted section), link text says "output aggregation"
- In Section 15.2 Accuracy benchmarking, link text says "data aggregation"

### LINKED ISSUE #267: Scrub document to ensure correct use of "should" and "may"  (Anne)
There are 12 occurrences of "should" in the document without it being "SHOULD".   Same goes for "may".

WilcoFiers:
- Closes #264
- Closes #257
- Closes #266
- Closes #239
- Closes #220
- Closes #266
- Closes #230 
- Closes #232
- Closes #267

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/pull/273.html" title="Last updated on Sep 24, 2018, 2:46 PM GMT (ff63f9f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/273/ef03048...ff63f9f.html" title="Last updated on Sep 24, 2018, 2:46 PM GMT (ff63f9f)">Diff</a>


moekraft:
@WilcoFiers I think we need to keep our edits atomic so we don't overwrite each other. I spent a good hour+ last week trying to resolve conflicts.

WilcoFiers:
Normally I'd agree. But if I was to do all of these separately I'd have 10 PRs with a lot of conflicts between them, and it would have been a lot harder for me to track my progress and see the end result. This just lets me move faster. This is all editorial stuff that the group has already agreed with so I don't intend this to be reviewed by the group. Everything that needs review from the group will go in its own PRs.

act-rules-format.bs
```diff
@@ -52,18 +52,21 @@ The separation between atomic rules and composite rules creates a division of re
...
-  <p>If any one of these rules passes, the table cell can pass the composite rule.</p>
+  <p>If any one of these rules passes, the test result of the composite rule for the table cell is "passed".</p>
 </div>
 
 
 ACT Rule Structure {#act-rule-structure}
 ===============================
 
-An ACT Rule MUST consist of the following items
+ACT Rules MUST be written in an format that conforms to the Web Content Accessibility Guidelines [[WCAG]] or a comperable accessibility guidelines recommendation. ACT test cases are allowed to contain inaccessible content. If any test case contains accessibility issues listed in [WCAG 2.1 Section 5.2.5 Non-Interference](https://www.w3.org/TR/WCAG21/#cc5), users MUST be warned of this in advance. Using an accessibility format supports people with disabilities. It also makes internationalisation of ACT rules easier.
```
annethyme: I am not sure how to understand this section. 
When you say "written in a format..." I am thinking it should be the ACT Rules Format they conform to.
Is the intent of this section that the presentation of the ACT rule should be accessible itself?

PR #276, opened 2018-09-24
Update accessibility requirements section

### LINKED ISSUE #227: clarifying the relationship between rule application and accessibility conformance
I agree with the note in the draft which suggests that using the terms "pass" and "fail" with respect both to individual conformance testing rules and to accessibility conformance itself is confusing. In particular, "passing" an accessibility conformance testing rule may result in "failing" to conform to an accessibility requirement asserted in WCAG or elsewhere. "Failing" an accessibility conformance testing rule may contribute to "passing" an accessibility conformance requirement.

I suggest avoiding the terms "pass" and "fail" in connection with ACT rules by using terms such as "satisfying" and "not satisfying" a rule. A rule is "satisfied" or "not satisfied" (alternatively, if you prefer, "unsatisfied") - never "passed" or "failed". Then we can say, without contradiction, that "satisfying" an ACT rule may result in "failing" to meet an accessibility conformance requirement (e.g., a success criterion) - and so forth. Such an explanation in these terms could usefully be included in the text.

### LINKED ISSUE #231: 7. Accessibility Requirements - Improvement suggestion
Change "... from different sources, such as local laws, internal standards, etc. "
to "... from different sources, such as **national or** local laws, internal standards, etc."

Accessibility legislation is usally on national or state level. Local laws are not that common.

### LINKED ISSUE #235: Outcomes of atomic tests
In the current approach atomic tests can have two types of outcome.

1. Outcomes that are used as input to composed rules and should not be interpreted as "The content passes or fails an accessibility reuqirement."

_[At least this is how I understand the last paragrah of section 7 "Outcomes from an ACT Rule SHOULD be consistent with the accessibility requirement, ..." - which is a bit confusing by the way.]_

2. Outcomes that need no further aggregation and can be used directly to make a statement about an accessibility reuqirements.

3. Maybe there are even atomic rules which can be both?

In my opinion this distinction should be made more explicit. This could be done by restricting the use of the terms "pass" and "fail" to the second type of outcome. Or by adding composed rules as "wrapper" to all atomic rule, i.e. adding composed rules even if only one atomic rule is involved. Or by adding this information to the rule structure (see also #232).

This issue is also related to your question in Section 11: In my opinion "test definition" and "aggregation definition" should not be merged.

### LINKED ISSUE #250: Accessibility Requirements: Updates requested (Shadi)
Editor’s note: The ACT Task Force is looking for feedback about the use of the term "pass" in relation to rules. While rules can "pass", their corresponding accessibility requirement can fail. A section has been added to make this explicit in the rule, but we would like to know if this is sufficient. 
Accessibility requirements are just that: A requirement that content must conform to for it to be considered accessible. These are usually WCAG success criteria but can include other requirements too. For example, an organization might have additional requirements, such as local laws or internal standards. These too are considered accessibility requirements and can be tested using ACT Rules. What the precise requirements are for any particular test is beyond the scope of the ACT Rules Format.
ACT Rules SHOULD identify the accessibility requirements that are not met when the outcome of a rule is Fail. An ACT Rule is a complete or partial test for one or more accessibility requirements. However, an ACT rule MUST NOT test more than the accessibility requirements it lists.
ACT Rules MUST indicate when they can not be used to determine that the accessibility requirement passed.
Outcomes from an ACT Rule SHOULD be consistent with the accessibility requirement, e.g. a rule only returns the outcome Fail when the content does not meet the accessibility requirement. This means that the rule maps to the accessibility requirement, as opposed to it merely being related to the requirement, thematically or otherwise. Because of this, atomic rules used in composed rules often do not map to any accessibility requirement. Failing the composed rule fails the accessibility requirement, but failing any of its atomic rules may not. In such cases the atomic rules MUST NOT list the accessibility requirements. These could be provided as background information instead.

Comments:

1.  An example is needed for this statement, _ACT Rules MUST indicate when they can not be used to determine that the accessibility requirement passed._

2. Is this a rule or precondition? _. In such cases the atomic rules MUST NOT list the accessibility requirements. 
[Accessibility Conformance Testing.docx](https://github.com/w3c/wcag-act/files/2293796/Accessibility.Conformance.Testing.docx)
_

WilcoFiers:
- Closes #231
- Closes #227
- Closes #235
- Closes #250

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/pull/276.html" title="Last updated on Oct 4, 2018, 5:01 PM GMT (437079d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/276/f6210f2...437079d.html" title="Last updated on Oct 4, 2018, 5:01 PM GMT (437079d)">Diff</a>

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -89,17 +89,40 @@ An ACT Rule MUST have a description that is in plain language and provides a bri
 Accessibility Requirements {#structure-accessibility-requirements}
 ==================================================================
 
-<div class=note>
-  **Editor's note:** The ACT Task Force is looking for feedback about the use of the term "pass" in relation to rules. While rules can "pass", their corresponding accessibility requirement can fail. A section has been added to make this explicit in the rule, but we would like to know if this is sufficient.
+Accessibility requirements are just that: A requirement that a particular web page must conform to for it to be considered accessible. This can (and usually does) include the WCAG success criteria. Often organizations have additional requirements which may come from different sources, such as laws, internal standards, etc. These too are considered accessibility requirements and can be tested using the ACT Rules Format. What the precise requirements are for any test is beyond the scope of the ACT Rules Format.
+
+[=Atomic rules=] and [=composite rules=] SHOULD identify the accessibility requirements that are *not satisfied* when the outcome of a rule is Fail. An ACT Rule is a complete or partial test for one or more accessibility requirements. This means that most ACT rules test only part of an accessibility requirement, but it MUST NOT test more than the accessibility requirement it lists.
+
+Because ACT Rules often have a smaller scope than the accessibility requirement they test, passing a rule does not necessarily mean that the accessibility requirement is *satisfied*. ACT Rules MUST indicate whether or not they can be used to determine that the accessibility requirement is *satisfied*.
+
+<div class=example>
+  <p>The following rule could **not** be used to satisfy [WCAG 2.1 Success criterion 2.4.2 Page Titled](https://www.w3.org/TR/WCAG21/#page-titled)</p>
+  <blockquote>
+    <p>This rule tests that an HTML page has a `title` element.</p>
+  </blockquote>
 </div>
 
-Accessibility requirements are just that: A requirement that a particular web page must conform to for it to be considered accessible. This can (and usually does) include the WCAG success criteria. Often organizations have additional requirements which may come from different sources, such as local laws, internal standards, etc. These too are considered accessibility requirements and can be tested using the ACT Rules Format. What the precise requirements are for any test is beyond the scope of the ACT Rules Format.
+<div class=example>
+  <p>The following rule could be used to satisfy [WCAG 2.1 Success criterion 2.4.2 Page Titled](https://www.w3.org/TR/WCAG21/#page-titled)</p>
+  <blockquote>
+    <p>This rule tests that the first `title` element of an HTML page describes the topic or purpose of that page.</p>
+  </blockquote>
+</div>
 
-[=Atomic rules=] and [=composite rules=] SHOULD identify the accessibility requirements that fail when the outcome of a rule is Fail. An ACT Rule is a complete or partial test for one or more accessibility requirements. This means that most ACT rules test only part of an accessibility requirement, but it MUST NOT test more than the accessibility requirement it lists.
+Outcomes from an ACT Rule MUST be consistent with the accessibility requirement, e.g. a rule only returns the outcome Fail when the content does *not satisfy* the accessibility requirement. If an ACT Rule indicates it can be used to satisfy the accessibility requirement, a Pass or Inapplicable result for that rule MUST satisfy the accessibility requirement.
 
-Because ACT Rules often have a smaller scope than the accessibility requirement they test, passing a rule does not necessarily mean that the accessibility requirement has passed. ACT Rules MUST indicate when they can not be used to determine that the accessibility requirement passed.
+This means that the rule maps to the accessibility requirement, as opposed to it merely being related to the requirement, thematically or otherwise. Because of this, atomic rules used in composite rules often do not map to any accessibility requirement. Failing the composite rule fails the accessibility requirement, but failing any of its atomic rules may not. In such cases the atomic rules MUST NOT list the accessibility requirements. These could be provided as background information instead.
 
-Outcomes from an ACT Rule SHOULD be consistent with the accessibility requirement, e.g. a rule only returns the outcome Fail when the content fails the accessibility requirement. This means that the rule maps to the accessibility requirement, as opposed to it merely being related to the requirement, thematically or otherwise. Because of this, atomic rules used in composite rules often do not map to any accessibility requirement. Failing the composite rule fails the accessibility requirement, but failing any of its atomic rules may not. In such cases the atomic rules MUST NOT list the accessibility requirements. These could be provided as background information instead.
+<div class=note>
+  There are two reasons why an ACT Rule might not list any accessibility requirements:
+
+  1. The ACT Rule is a best practice that is not required by any accessibility standard
+  2. The ACT Rule is an [=atomic rule=] and its result is used in a [=composite rule=] that does map to an accessibility requirement. See: [ACT Rule Types](#rule-types)
+</div>
+
+<div class=note>
+  In WCAG 2, success criteria do not evaluate to Pass, Fail or Inapplicable. Rather they can be *satisfied* (or not). (See the [WCAG 2.1 definition: satisfies a success criterion](https://www.w3.org/TR/WCAG21/#dfn-satisfies).) A Fail result from an ACT Rule means the success criteria listed as "accessibility requirements" were not satisfied for the particular page. A page may still conform to WCAG 2 if a conforming alternative version is provided, as described in [WCAG 2.1 Conformance Requirement 1](https://www.w3.org/TR/WCAG21/#cc1).
+</div>
 
 
 Aspects Under Test (Atomic rules only) {#input-aspects}
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -86,43 +86,30 @@ Rule Description {#structure-description}
 An ACT Rule MUST have a description that is in plain language and provides a brief explanation of what the rule does.
 
 
-Accessibility Requirements {#structure-accessibility-requirements}
+Accessibility Requirements {#accessibility-requirements}
 ==================================================================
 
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -89,17 +89,40 @@ An ACT Rule MUST have a description that is in plain language and provides a bri
 Accessibility Requirements {#structure-accessibility-requirements}
 ==================================================================
 
-<div class=note>
-  **Editor's note:** The ACT Task Force is looking for feedback about the use of the term "pass" in relation to rules. While rules can "pass", their corresponding accessibility requirement can fail. A section has been added to make this explicit in the rule, but we would like to know if this is sufficient.
+Accessibility requirements are just that: A requirement that a particular web page must conform to for it to be considered accessible. This can (and usually does) include the WCAG success criteria. Often organizations have additional requirements which may come from different sources, such as laws, internal standards, etc. These too are considered accessibility requirements and can be tested using the ACT Rules Format. What the precise requirements are for any test is beyond the scope of the ACT Rules Format.
```
nitedog: - "a particular web page" -> "particular content" (to make it less page-centric, for example documents, web or mobile applications, etc.)
- "Often" -> "Some" (I think "often" is implying a level of frequency that is hard to prove, whereas "some" is more open)
- remove "which may come" (to get rid of "may" keyword)
- "are considered" -> "can be considered" (qualifier needed IMO)

act-rules-format.bs
```diff
@@ -89,17 +89,40 @@ An ACT Rule MUST have a description that is in plain language and provides a bri
...
+
+[=Atomic rules=] and [=composite rules=] SHOULD identify the accessibility requirements that are *not satisfied* when the outcome of a rule is Fail. An ACT Rule is a complete or partial test for one or more accessibility requirements. This means that most ACT rules test only part of an accessibility requirement, but it MUST NOT test more than the accessibility requirement it lists.
+
```
nitedog: Editorial suggestion: "An ACT Rule is a complete or partial test for one or more accessibility requirements but most ACT Rules test only part of an accessibility requirement. However, an ACT Rule MUST NOT test more than the accessibility requirement it lists." (to me the current text is confusing because it mixes different thoughts without clear separation between them)

act-rules-format.bs
```diff
@@ -89,17 +89,40 @@ An ACT Rule MUST have a description that is in plain language and provides a bri
...
+Because ACT Rules often have a smaller scope than the accessibility requirement they test, passing a rule does not necessarily mean that the accessibility requirement is *satisfied*. ACT Rules MUST indicate whether or not they can be used to determine that the accessibility requirement is *satisfied*.
```
nitedog: How do they indicate that? This needs to be specified.

act-rules-format.bs
```diff
@@ -89,17 +89,40 @@ An ACT Rule MUST have a description that is in plain language and provides a bri
...
+<div class=example>
+  <p>The following rule could be used to satisfy [WCAG 2.1 Success criterion 2.4.2 Page Titled](https://www.w3.org/TR/WCAG21/#page-titled)</p>
+  <blockquote>
+    <p>This rule tests that the first `title` element of an HTML page describes the topic or purpose of that page.</p>
+  </blockquote>
+</div>
+
+Outcomes from an ACT Rule MUST be consistent with the accessibility requirement, e.g. a rule only returns the outcome Fail when the content does *not satisfy* the accessibility requirement. If an ACT Rule indicates it can be used to satisfy the accessibility requirement, a Pass or Inapplicable result for that rule MUST satisfy the accessibility requirement.
+
+This means that the rule maps to the accessibility requirement, as opposed to it merely being related to the requirement, thematically or otherwise. Because of this, atomic rules used in composite rules often do not map to any accessibility requirement. Failing the composite rule fails the accessibility requirement, but failing any of its atomic rules may not. In such cases the atomic rules MUST NOT list the accessibility requirements. These could be provided as background information instead.
```
nitedog: IMO this part is quite confusing: "MUST be consistent" only applies to Fail or, if the rule indicates it applies to the entire requirement, also for Pass. Inapplicable is an exception anyway because WCAG doesn't define that. But if it is an atomic rule then also the Fail can be inconsistent, and you have to look at it in context of the composite rule.

Not sure what to do about this but it seems we are trying to mix *two* outputs rather than one: (1) the outcome of the rule, and (2) any implications this outcome could have on meeting the accessibility requirement. Maybe we need to have two such outputs? The outcome of the rule is *always* returned, and the other is conditional and must be consistent with the accessibility requirement.

act-rules-format.bs
```diff
@@ -89,17 +89,40 @@ An ACT Rule MUST have a description that is in plain language and provides a bri
...
+  <blockquote>
+    <p>This rule tests that an HTML page has a `title` element.</p>
+  </blockquote>
 </div>
 
-Accessibility requirements are just that: A requirement that a particular web page must conform to for it to be considered accessible. This can (and usually does) include the WCAG success criteria. Often organizations have additional requirements which may come from different sources, such as local laws, internal standards, etc. These too are considered accessibility requirements and can be tested using the ACT Rules Format. What the precise requirements are for any test is beyond the scope of the ACT Rules Format.
+<div class=example>
+  <p>The following rule could be used to satisfy [WCAG 2.1 Success criterion 2.4.2 Page Titled](https://www.w3.org/TR/WCAG21/#page-titled)</p>
+  <blockquote>
+    <p>This rule tests that the first `title` element of an HTML page describes the topic or purpose of that page.</p>
```
annethyme: When comparing this to actual rules, this looks like a rule that only applies to documents that have a `<title>` element, and since non-applicable = pass in WCAG, I find this example a little confusing, since it's a bit unclear if it actually tests for the whole requirement.

I would suggest changing it to something along these lines:

> This rule tests that that an HTML page has a `title` element, and that the first `title` element describes the topic or purpose of that page

act-rules-format.bs
```diff
@@ -89,17 +89,40 @@ An ACT Rule MUST have a description that is in plain language and provides a bri
...
+  </blockquote>
+</div>
+
+Outcomes from an ACT Rule MUST be consistent with the accessibility requirement, e.g. a rule only returns the outcome Fail when the content does *not satisfy* the accessibility requirement. If an ACT Rule indicates it can be used to satisfy the accessibility requirement, a Pass or Inapplicable result for that rule MUST satisfy the accessibility requirement.
```
annethyme: I find this wording a bit ambiguous: "a rule only returns the outcome Fail when the content does *not satisfy* the accessibility requirement" - as I read it this could mean both "failing the requirement" or "cannot be determined to satisfy the requirement".

I suggest something along these lines:

> "a rule only returns the outcome Fail when the content has been determined to *not satisfy* the accessibility requirement"

act-rules-format.bs
```diff
@@ -89,17 +89,40 @@ An ACT Rule MUST have a description that is in plain language and provides a bri
...
+<div class=note>
+  In WCAG 2, success criteria do not evaluate to Pass, Fail or Inapplicable. Rather they can be *satisfied* (or not). (See the [WCAG 2.1 definition: satisfies a success criterion](https://www.w3.org/TR/WCAG21/#dfn-satisfies).) A Fail result from an ACT Rule means the success criteria listed as "accessibility requirements" were not satisfied for the particular page. A page may still conform to WCAG 2 if a conforming alternative version is provided, as described in [WCAG 2.1 Conformance Requirement 1](https://www.w3.org/TR/WCAG21/#cc1).
```
annethyme: Should there also be a note about "Inapplicable" mapping to "satisfied" in WCAG 2?

annethyme: In [rule aggregation](https://pr-preview.s3.amazonaws.com/w3c/wcag-act/pull/276.html#rule-aggregation) we write "Example: An expert evaluates a success criterion to fail on a specific page." This needs to be updated to match this new definition.

act-rules-format.bs
annethyme: Isn't "Failed" / "Passed" / "Inapplicable" the agreed upon terminology for outcomes of rules? 
In general af Ctrl+F reveals a lot of "Fail" and "Pass" throughout the document.

act-rules-format.bs
tbostic32: There is currently this text in an editors note in the published copy:
"While rules can "pass", their corresponding accessibility requirement can fail. A section has been added to make this explicit in the rule, but we would like to know if this is sufficient."

I am not sure which section we are referring to here, so please let me know.
Seems like some small modifications to the accessibility requirements section could fix this.

act-rules-format.bs
maryjom: I don't think "particular" needs to be in there either, so it could simply say "content".  

I agree with your other comments on this part.

act-rules-format.bs
maryjom: I think we should show in the examples what part of the rule (what section) that these statements should be made in.

maryjom:
One question: Should best practices be marked as such in the rule rather than leaving the accessibility requirements blank?

WilcoFiers:
PR superseded by #283

Issue #272, opened 2018-09-20
Clear split between requirements for rule authors vs. rule implementers


annethyme:
These sections are as the only ones in the ACT Rules Format related to the **implementation** of rules rather than the ACT rules themselves: 
* [14. ACT Data Format (Output Data)](https://www.w3.org/TR/act-rules-format/#output)
* [15.2. Accuracy Benchmarking](https://www.w3.org/TR/act-rules-format/#quality-accuracy) - AT RISK?
* [16. Rule Aggregation](https://www.w3.org/TR/act-rules-format/#output-aggregation)

Mixed in with these are these sections that are actually requirements for ACT rules themselves: 
* [15.1. Test Cases (Atomic rules only)](https://www.w3.org/TR/act-rules-format/#quality-test-cases)
* [17.1. Change Log](https://www.w3.org/TR/act-rules-format/#quality-updates-changes)
* [17.2. Issues List](https://www.w3.org/TR/act-rules-format/#quality-updates-issues)

These three requirements for ACT Rules are however not listed in the [4. ACT Rule Structure](https://www.w3.org/TR/act-rules-format/#act-rule-structure), making them seem less of a requirement for ACT Rules.

I suggest we:
* somehow make it clear from the structure of the document and in the wording of the sections which sections are specifically **implementation** related
* Add Test Cases, Change log and Issues list to the [4. ACT Rule Structure](https://www.w3.org/TR/act-rules-format/#act-rule-structure)

annethyme:
@WilcoFiers, you mentioned that this would go on the agenda this Thursday, but I don't see it explicitly. Do you see it as covered by the items on the agenda, https://github.com/w3c/wcag-act/pull/277 or https://github.com/w3c/wcag-act/pull/276?

PR #277, opened 2018-09-24
Update rule accuracy section

### LINKED ISSUE #236: 15.2. Accuracy Benchmarking - Update definition
Your definition of accuracy is not aligned with the definition used in statistics / information science. In general, higher accuracy means that an approach is better. In your definition a higher number means that the approach performed worse!

See [Wikipedia - Accuracy and precision:](https://en.wikipedia.org/wiki/Accuracy_and_precision#In_binary_classification)
"Accuracy is the proportion of true results (both true positives and true negatives) among the total number of cases examined."

Another useful approach might be to consider **precision and recall** as indicators in the benchmarking.
See [Wikipedia - Precision and recall:](https://en.wikipedia.org/wiki/Precision_and_recall)
"Precision is the fraction of relevant instances among the retrieved instances, while recall is the fraction of relevant instances that have been retrieved over the total amount of relevant instances."

### LINKED ISSUE #239: False negatives definition

### LINKED ISSUE #251: Test cases for composed rules
The proposal currently suggests that atomic rules require test cases, but that composed rules do not. In writing composed rules, I found myself copying over some of the test cases from atomic rules to ensure that all the atomic rules are aggregated correctly. I think even for composed rules we should have test cases.

WilcoFiers:
- Closes #251
- Closes #236
- Closes #239

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/pull/277.html" title="Last updated on Oct 4, 2018, 10:44 AM GMT (d145276)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/277/f6210f2...d145276.html" title="Last updated on Oct 4, 2018, 10:44 AM GMT (d145276)">Diff</a>

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -326,7 +326,7 @@ When executing a test, the test aspect(s), for instance an HTML code snippet, is
 Rule Accuracy {#accuracy}
 -------------------------
 
-This section is non-normative.
+This section is *non-normative*.
 
 The web is ever changing, and technologies are used in such diverse and creative ways that it is impossible to predict in advance, all the ways that accessibility issues can occur and all the ways they can be solved for. When writing ACT Rules, it is almost inevitable that edge cases will be overlooked during the design of a rule, or that new technologies will emerge that introduce exceptions.
 
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -315,31 +315,28 @@ The definition of a rule MUST always result in one of the following outcomes:
 </div>
 
 
-Rule Quality Assurance {#quality}
-=================================
+Test Cases {#test-cases}
+========================
 
-Test Cases (Atomic rules only) {#quality-test-cases}
-----------------------------------------------------
-
-Test cases are (snippets of) content that can be used to validate the implementation of an [=atomic rule=]. They consist of two pieces of data, a snippet of each [test aspect](#input-aspects) for a rule, and the [expected result](#output) that should come from that rule. Test cases serve two functions, firstly as example scenarios for readers to understand when a rule passes, when it fails, and when it is inapplicable. But also for developers and users of automated accessibility test tools to validate that a rule is correctly implemented.
+Test cases are (snippets of) content that can be used to validate the implementation of an ACT Rules. Each rule MUST have tefst cases for passed, failed and inapplicable scenarios. A test case consist of two pieces of data, a snippet of each [test aspect](#input-aspects) for a rule, and the [expected result](#output) that should come from that rule. Test cases serve two functions, firstly as example scenarios for readers to understand when a rule passes, when it fails, and when it is inapplicable. But also for developers and users of automated accessibility test tools to validate that a rule is correctly implemented.
 
 When executing a test, the test aspect(s), for instance an HTML code snippet, is evaluated by applying the rule's [applicability](#applicability) and [expectations](#expectations). The result is then compared to the expected result of the test case. The expected result consists of a list of [test targets](#output-test-target) and the expected [outcome](#output-outcome) (Passed, Failed, Inapplicable) of the evaluation.
 
 
-Accuracy Benchmarking {#quality-accuracy}
------------------------------------------
+Rule Accuracy {#accuracy}
+-------------------------
 
-The web is ever changing, and technologies are used in such diverse and creative ways that it is impossible to predict in advance, all the ways that accessibility issues can occur and all the ways they can be solved for. When writing ACT Rules, it is almost inevitable that exceptions will be overlooked during the design of a rule, or that new technologies will emerge that introduce new exceptions.
+This section is non-normative.
 
-This makes it important to be able to regularly test if the rule has the accuracy that is expected of it. This can be done by benchmark testing. In benchmark testing, the accuracy of a rule is measured by comparing its results to those obtained through accessibility expert testing.
+The web is ever changing, and technologies are used in such diverse and creative ways that it is impossible to predict in advance, all the ways that accessibility issues can occur and all the ways they can be solved for. When writing ACT Rules, it is almost inevitable that edge cases will be overlooked during the design of a rule, or that new technologies will emerge that introduce exceptions.
 
-The accuracy of a rule is the average between the false positives and false negatives, which are in turn calculated as follows:
+When a test result incorrectly indicates non-conformance to an accessibility requirement, this is known as a false positive. Opposite, when a rule incorrectly indicates conformance, this is a false negative. A percentage of false positives and false negatives can be calculated by comparing it to results from expert accessibility evaluations:
 
-- **False positives**: This is the percentage of [test targets](#output-test-target), that were failed by the rule, but were not failed by an accessibility expert.
+- **False positives**: This is the percentage of [test targets](#output-test-target), that were failed by the rule, but were conformanct to the [accessibility requirements](#structure-accessibility-requirements).
 
-- **False negatives**: This is the percentage of test targets, that were passed by the rule, but were failed by an accessibility expert.
+- **False negatives**: This is the percentage of test targets, that were passed by the rule, but were non-conformant to the [accessibility requirements](#structure-accessibility-requirements).
 
-There are several ways this can be done. For instance, accessibility test tools can implement a feature which lets users indicate that a result is in error, or pages that for which accessibility results are known, can be tested using ATT, and the results are compared. To compare results from ACT Rules to those of expert evaluations, [data aggregation](#output-aggregation) may be necessary.
+The possibility of false positives and false negatives with ACT rules makes it important to have a process in place for designing and maintaining accurate rules. An example of such a process is the [ACT Review Process](https://w3c.github.io/wcag-act-rules/review-process.html). 
 
 
 Rule Aggregation {#rule-aggregation}
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -73,6 +73,7 @@ An ACT Rule MUST consist of the following items.
 * [Expectations](#expectations)
 * [Limitations, Assumptions or Exceptions](#structure-limitations-assumptions-exceptions), if any
 * [Accessibility Support](#acc-support) (optional)
+* [Test Cases](#test-cases)
 
 Rule Identifier {#rule-identifier}
 ==================================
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -315,31 +315,28 @@ The definition of a rule MUST always result in one of the following outcomes:
...
-
-Test cases are (snippets of) content that can be used to validate the implementation of an [=atomic rule=]. They consist of two pieces of data, a snippet of each [test aspect](#input-aspects) for a rule, and the [expected result](#output) that should come from that rule. Test cases serve two functions, firstly as example scenarios for readers to understand when a rule passes, when it fails, and when it is inapplicable. But also for developers and users of automated accessibility test tools to validate that a rule is correctly implemented.
+Test cases are (snippets of) content that can be used to validate the implementation of an ACT Rules. Each rule MUST have tefst cases for passed, failed and inapplicable scenarios. A test case consist of two pieces of data, a snippet of each [test aspect](#input-aspects) for a rule, and the [expected result](#output) that should come from that rule. Test cases serve two functions, firstly as example scenarios for readers to understand when a rule passes, when it fails, and when it is inapplicable. But also for developers and users of automated accessibility test tools to validate that a rule is correctly implemented.
```
nitedog: - tefst -> test
- passed, failed and inapplicable -> passed, failed, and inapplicable (oxford comma)
- "that should come from" -> "from applying" (to remove "should")
- "But also for" -> "It also serves" (personal preference)
- "automated accessibility test tools" -> "accessibility testing tools and methodologies" (because this also applies to manual rules)

act-rules-format.bs
```diff
@@ -315,31 +315,28 @@ The definition of a rule MUST always result in one of the following outcomes:
...
+Rule Accuracy {#accuracy}
+-------------------------
 
-The web is ever changing, and technologies are used in such diverse and creative ways that it is impossible to predict in advance, all the ways that accessibility issues can occur and all the ways they can be solved for. When writing ACT Rules, it is almost inevitable that exceptions will be overlooked during the design of a rule, or that new technologies will emerge that introduce new exceptions.
+This section is *non-normative*.
 
-This makes it important to be able to regularly test if the rule has the accuracy that is expected of it. This can be done by benchmark testing. In benchmark testing, the accuracy of a rule is measured by comparing its results to those obtained through accessibility expert testing.
+The web is ever changing, and technologies are used in such diverse and creative ways that it is impossible to predict in advance, all the ways that accessibility issues can occur and all the ways they can be solved for. When writing ACT Rules, it is almost inevitable that edge cases will be overlooked during the design of a rule, or that new technologies will emerge that introduce exceptions.
```
nitedog: - "solved for" -> "checked" (copyedit suggestion)
- "new technologies will emerge" -> "technologies will emerge or change" (personal knit)

act-rules-format.bs
```diff
@@ -315,31 +315,28 @@ The definition of a rule MUST always result in one of the following outcomes:
...
 
-The accuracy of a rule is the average between the false positives and false negatives, which are in turn calculated as follows:
+When a test result incorrectly indicates non-conformance to an accessibility requirement, this is known as a false positive. Opposite, when a rule incorrectly indicates conformance, this is a false negative. A percentage of false positives and false negatives can be calculated by comparing it to results from expert accessibility evaluations:
 
-- **False positives**: This is the percentage of [test targets](#output-test-target), that were failed by the rule, but were not failed by an accessibility expert.
+- **False positives**: This is the percentage of [test targets](#output-test-target), that were failed by the rule, but were conformanct to the [accessibility requirements](#structure-accessibility-requirements).
 
-- **False negatives**: This is the percentage of test targets, that were passed by the rule, but were failed by an accessibility expert.
+- **False negatives**: This is the percentage of test targets, that were passed by the rule, but were non-conformant to the [accessibility requirements](#structure-accessibility-requirements).
```
nitedog: Same as above
- "were conformant" -> "conform" (suggestion)
- do we need to add something like "measured by the test cases provided" or such to indicate how conformance is measured?

act-rules-format.bs
```diff
@@ -315,31 +315,28 @@ The definition of a rule MUST always result in one of the following outcomes:
...
+This section is *non-normative*.
 
-This makes it important to be able to regularly test if the rule has the accuracy that is expected of it. This can be done by benchmark testing. In benchmark testing, the accuracy of a rule is measured by comparing its results to those obtained through accessibility expert testing.
+The web is ever changing, and technologies are used in such diverse and creative ways that it is impossible to predict in advance, all the ways that accessibility issues can occur and all the ways they can be solved for. When writing ACT Rules, it is almost inevitable that edge cases will be overlooked during the design of a rule, or that new technologies will emerge that introduce exceptions.
 
-The accuracy of a rule is the average between the false positives and false negatives, which are in turn calculated as follows:
+When a test result incorrectly indicates non-conformance to an accessibility requirement, this is known as a false positive. Opposite, when a rule incorrectly indicates conformance, this is a false negative. A percentage of false positives and false negatives can be calculated by comparing it to results from expert accessibility evaluations:
```
annethyme: We talk about test results of rules here. 
Do we mean _implementations_ of rules here or are we trying to infer anything about _the rules themselves_ from the implementations? I am not even sure myself if this section applies to one or the other.

I guess that if the implementations are showing "correct" results for test cases, but still differences to expert evaluations, that would be an indicator that the rule itself has false postives/negatives.

act-rules-format.bs
```diff
@@ -315,31 +315,28 @@ The definition of a rule MUST always result in one of the following outcomes:
...
 
-- **False negatives**: This is the percentage of test targets, that were passed by the rule, but were failed by an accessibility expert.
+- **False negatives**: This is the percentage of test targets, that were passed by the rule, but were non-conformant to the [accessibility requirements](#structure-accessibility-requirements).
 
-There are several ways this can be done. For instance, accessibility test tools can implement a feature which lets users indicate that a result is in error, or pages that for which accessibility results are known, can be tested using ATT, and the results are compared. To compare results from ACT Rules to those of expert evaluations, [data aggregation](#output-aggregation) may be necessary.
+The possibility of false positives and false negatives with ACT rules makes it important to have a process in place for designing and maintaining accurate rules. An example of such a process is the [ACT Review Process](https://w3c.github.io/wcag-act-rules/review-process.html).
```
annethyme: As per @nitedog's comments about adding "measured by the test cases provided", I think that we need to separately treat two different types of false positives/negatives here: 
* When an implementation has false positives/negatives when measuring against the test cases provided for an ACT rule
* When implementations of a rule live up to the test cases provided for a rule, but still has false positives/negatives compared to expert testing, this could indicate that the rule itself is inaccurate or that the test cases provided are not comprehensive - and THEN it is important with the process

To me it seems that the current section is a bit of a mix of the two, or at least leaves too much room to interpret it in both directions.

act-rules-format.bs
annethyme: "of an ACT Rules" --> "of ACT Rules" OR "of an ACT Rule" 

Btw, do we write ACT Rules or ACT rules? Both is used in the spec, and I am not sure if it is according to a pattern.

act-rules-format.bs
```diff
@@ -315,31 +315,28 @@ The definition of a rule MUST always result in one of the following outcomes:
...
+Test cases are (snippets of) content that can be used to validate the implementation of an ACT Rules. Each rule MUST have tefst cases for passed, failed and inapplicable scenarios. A test case consist of two pieces of data, a snippet of each [test aspect](#input-aspects) for a rule, and the [expected result](#output) that should come from that rule. Test cases serve two functions, firstly as example scenarios for readers to understand when a rule passes, when it fails, and when it is inapplicable. But also for developers and users of automated accessibility test tools to validate that a rule is correctly implemented.
 
 When executing a test, the test aspect(s), for instance an HTML code snippet, is evaluated by applying the rule's [applicability](#applicability) and [expectations](#expectations). The result is then compared to the expected result of the test case. The expected result consists of a list of [test targets](#output-test-target) and the expected [outcome](#output-outcome) (Passed, Failed, Inapplicable) of the evaluation.
 
 
-Accuracy Benchmarking {#quality-accuracy}
------------------------------------------
+Rule Accuracy {#accuracy}
```
annethyme: Should this really be a heading under "Test Cases"?

act-rules-format.bs
```diff
@@ -315,31 +315,28 @@ The definition of a rule MUST always result in one of the following outcomes:
 </div>
 
 
-Rule Quality Assurance {#quality}
-=================================
+Test Cases {#test-cases}
```
annethyme: Should this be linked from the [ACT Rule Structure](https://pr-preview.s3.amazonaws.com/w3c/wcag-act/pull/276.html#act-rule-structure)? I raised an issue on this and related questions last week: https://github.com/w3c/wcag-act/issues/272

act-rules-format.bs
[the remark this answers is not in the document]
tbostic32: Could we say:
"A percentage of false positives and false negatives can be estimated by comparing it to a set of known test cases."

I am thinking that calibrated test cases = known test cases, wondering if known is more readable? We can discuss. 
Also wondering if estimated is a better term for this calculation since we are doing this in a known environment, but that doesn't mean the rule will perform as accurately "in the wild".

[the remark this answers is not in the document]
nitedog: Agree with "known" instead of "calibrated". But regarding your second point, maybe say "known results" to not limit this to test cases?

tbostic32: Have we flipped the definition of false positive and false negative here? I think of conformance as being positive and non-conformance as being negative. Thus if you incorrectly indicate conformance it would be a false-positive and incorrectly indicating non-conformance would be a false negative.

act-rules-format.bs
maryjom: Also in that Test Cases section. The third sentence should read: A test case "consists" of two pieces of data...

act-rules-format.bs
maryjom: In addition to this, shouldn't the phrase "designing and maintaining accurate rules" include the testing of the rules?  You have to test the implementation of the rules to find out their accuracy.

WilcoFiers:
This PR is superseded by #283

Issue #224, opened 2018-07-05
List of features for exit critiera


WilcoFiers:
I went through the document and listed out all the things that we may want to demonstrate. This is a fairly long and complex list, we should discuss where / if / how we can trim this down:

General
  - All rules must conform to the ACT Rules Format

Accessibility requirements
  - At least one atomic rule that has a requirement in WCAG 2
  - At least one composite rule that has a requirement in WCAG 2
  - At least one atomic rule has no accessibility requirement, but is used in a composite rule that has a requirement in WCAG 2
  - At least one rule that has a accessibility requirement that is not part of WCAG

Aspects under test
  - At least two different aspects under tests are used
  - At least two different web content technologies

Applicability
  - At least one applicability is expressed with a formal selector syntax, in addition to a plain language description
  - At least one atomic rule has an applicability different from the composite rule it is used in

Expectations
  - At least one rule with one expectation
  - At least one rule with two expectations
  - At least one rule with more than two expectations

Limitations, Assumptions or Exceptions
  - At least one rules without Limitations, Assumptions or Expectations
  - At least one rule where interpretation is documented in Limitations Assumptions or Expectations

Accessibility Support
  - At least one rule where no accessibility support information was necessary
  - At least one rule where accessibility support information was included

Aggregation
  - At least one implementation that takes results from rules and aggregates it

Changelog
  - At least two rules that include a changelog

Issues list
  - At least two rules that include an issues list

Test cases
  - Every applicability must have at least one inapplicable test case
  - Every expectation must have at least one pass and one fail test case

Implementations and data format
  - At least two fully automated independent implementations of the same ACT rule
    (All test cases that should pass are passing, all test cases that should fail are failing)
  - At least two semi-automated independent implementations of the same ACT rule
    (Some results running against test cases are inconclusive, none of the results contradict the test cases)
  - At least two manual testing methodologies where one of their rules is comparable to an ACT rule

NOT INCLUDED:
  - Benchmarking, expect this to be removed

rdeltour:
* test cases for atomic rules
* accuracy benchmarks for some rules (?)

annethyme:
If it's not already covered by "Implementations", I suggest that we state how many different organisations we want to see implementations from in tools/methodologies.

cpandhi:
+1 to 

- X no. of  rules implementation (atomic and Composed) as per the ACT format

- test cases for the implementation

- Accuracy benchmark with expert review

moekraft:
[...]

Issue #216, opened 2018-06-03
Need non-versioned biblio reference tags


moekraft:
Work with bike shed team to generate non-versioned biblio tags for CSS, UAAG, HTTP and SVG.

moekraft:
UAAG: A non-versioned url exists, however, it opens the 1.0 spec, https://www.w3.org/TR/UAAG/ I am currently using [UAAG20] tag for the informative reference which links to the 2.0 spec, https://www.w3.org/TR/UAAG20.
CSS: A a non-versioned url is available, https://www.w3.org/TR/CSS/. However, I needed to use a specific version, [CSS2], in the tag, otherwise, I receive an error from bikeshed.
HTTP: I am using [HTTP11]. I don't believe there is a non-versioned url.
SVG: There is a non-versioned url, however, it opens the 1.1 spec, https://www.w3.org/TR/SVG/. I am using [SVG2], https://www.w3.org/TR/SVG2/.

WilcoFiers:
@moekraft what is the status on this?

moekraft:
@maryjom @nitedog  It does look like the WAI-ARIA tag is resolving to the the ARIA 1.0 and not the non-versioned TR, https://www.w3.org/TR/wai-aria/ So yes, WAI-ARIA should be added to the list for the Bikeshed team.

moekraft:
UAAG: A non-versioned url exists, however, it opens the 1.0 spec, https://www.w3.org/TR/UAAG/ Spec currently uses[UAAG20] tag for the informative reference which links to the 2.0 spec, https://www.w3.org/TR/UAAG20.
CSS: A a non-versioned url is available, https://www.w3.org/TR/CSS/. However, this receives an error from Bikeshed. Need to use a specific version, e.g. [CSS2].
HTTP: Spec uses [HTTP11]. I don't believe there is a non-versioned url.
SVG: There is a non-versioned url, however, it opens the 1.1 spec, https://www.w3.org/TR/SVG/. Spec uses [SVG2], https://www.w3.org/TR/SVG2/.
[WAI-ARIA] References WAI-ARIA 1.0. Should reference non-versioned TR, https://www.w3.org/TR/wai-aria/, which opens WAI-ARIA 1.1

moekraft:
Opened Bikeshed issue 1359: https://github.com/tabatkins/bikeshed/issues/1359

WilcoFiers:
@moekraft can you confirm this is done?

[...]

PR #283, opened 2018-10-08
CR Proposal

### LINKED ISSUE #162: Clarify the output format definition
Section [ACT Data Format (Output Data)](https://w3c.github.io/wcag-act/act-rules-format.html#output) says that a rule’s execution should 

> return a set of zero or more results each of which MUST have at least the following properties:
> - test (rule ID)
> - subject (Web page)
> - pointer (Test target)
> - outcome (Passed or Failed)

I find that this lacks clarity in several aspects:

- the `test` and `subject` properties are the same for all the test targets, do they have to be repeated in all the test targets’ individual results?
- does it mean that the test output should contain the individual results for **all** the individual tests targets, even when they pass? A single global `pass` outcome may be preferable over having potentially thousands of individual `pass` outcomes, one for each target).
- can we define the result in a more structured manner: a global outcome + a collection of individual outcomes for failing test target?
- the `outcome` is defined here as "passed or failed" where it later is described as accepting also the value "Inapplicable" (in 8.2).

Section [8.2 Outcome](https://w3c.github.io/wcag-act/act-rules-format.html#output-outcome) says:

> The procedure of a rule MUST always return one of the following outcomes

This apparently applies to the global outcome, but this concept is not explained clearly in the introduction of the "ACT Data Format" section.

Section [8.3 Ensure comparable results](https://w3c.github.io/wcag-act/act-rules-format.html#output-comparable) says:

> ACT Rule output data MUST include (…) the Web page, file or other test subject the rule was applied to

which is somewhat contradictory with the following section explaining that only a pointer will do.

Section [8.4 Test subject](https://w3c.github.io/wcag-act/act-rules-format.html#output-test-subject) is really about the `pointer` property of the outcome, so it should be made more explicit.

### LINKED ISSUE #163: Clarify how to do "Implementation validation" and "accuracy benchmarking" for manual or semi-automated rules
Section [9.1 Implementation validation](https://w3c.github.io/wcag-act/act-rules-format.html#quality-implement) says:

> Implementation tests are tests for accessibility test tools. These tests enable the developers and users of ATTs to validate the implementation of an ACT Rule.

Section [9.2 Accuracy benchmarking] says:

> In benchmark testing, the accuracy of a rule is measured, by comparing it’s results to those obtained through accessibility expert testing.

and

> For instance, accessibility test tools can implement a feature which lets users indicate that a result is in error, or pages that for which accessibility results are known, can be tested using ATT, and the results are compared. To compare results from ACT Rules to those of expert evaluations, data aggregation may be necessary.

This sounds like these sections only apply to automated rules, as implemented by ATT.
Given that our format also covers manual test rules, this should be clarified.

### LINKED ISSUE #165: Clarify the definition of Rule Aggregation
Section [9.3 Rule Aggregation](https://w3c.github.io/wcag-act/act-rules-format.html#output-aggregation) does not properly define the aggregation mechanism. In its current state it’s mostly a description of the issue. We should consider expanding this section to properly define how rule’s results should be aggregated.

Also, having this section in "Rule Quality Assurance" looks a bit odd.

### LINKED ISSUE #187: Examples need to be identified accessibly
The only visible markers of an example are a border colour which has a contrast ration o less than 2:1, and a background that is very very low contrast.

Adding the word "Example" to a section would be extremely helpful.

### LINKED ISSUE #221: Section “13. Accessibility Support” unclear
From [Stefan Schnabel](https://lists.w3.org/Archives/Public/public-wcag-act-comments/2018Jul/0000.html):

> Section “13. Accessibility Support”:
>  
> “Because of this, ACT Rules are not necessarily applicable in all test scenarios. For instance, a web page that has to work in assistive technologies that have no WAI-ARIA [WAI-ARIA] support, wouldn’t be tested with an ACT Rule that relies on WAI-ARIA support, since this could lead to false positive results.”
>  
> This is not understandable without further examples or rewording especially what “false positive results” would mean.
> 
> Concealment does not improve things.
> 
> I mean, shouldn’t the rule set actively *encourage* the USE of WAI-ARIA instead of doing protectionism for older AT and user agents?
> 
> Of course, respective rule messaging should be flagged accordingly.

### LINKED ISSUE #226: ACT i18n checks
For review: Do any of the questions on this page answer to yes:

https://www.w3.org/International/techniques/developing-specs

Please give me a Thumbs up on this topic if you think it doesn't, or leave a comment about what you think does.

### LINKED ISSUE #227: clarifying the relationship between rule application and accessibility conformance

### LINKED ISSUE #235: Outcomes of atomic tests

### LINKED ISSUE #236: 15.2. Accuracy Benchmarking - Update definition

### LINKED ISSUE #250: Accessibility Requirements: Updates requested (Shadi)

### LINKED ISSUE #252: 16. Rule Aggregation - Add Requirement
It is not clear what the requirements are in this section. At least there is this implied requirement should be included in the spec:

"Aggregation of results from ACT Rules MUST be aligned with the accessibility requirements they refer to".

Are there other requirements that rule writers and implementers need to adhere to?

### LINKED ISSUE #261: Require Rule ID's in Atomic Rule List
Suggest including that Rule Identifiers and Rule Descriptions are required in the Composite rules' Atomic Rule List under section 9.

Agree with [@cpandhi about Example 5](https://github.com/w3c/wcag-act/issues/241#issuecomment-420722141) under section 11.2.

With rule identifiers, Example 5 could be:
For each test target, the outcome of one of the following rules is passed :
Video elements have an audio description (video-audio-desc)
Video elements have a text alternative (video-text-alt)

### LINKED ISSUE #271: Need a definition of Test Target
From Shadi, 

14.1. Test Subject
When a single URL can be used to reference the web page, or other test subject, this URL MUST be used. In scenarios where more complex actions are required to obtain the test subject (in the state that it is to be tested), it is left to ATT developers to determine which method is best used to express the test subject.

This section is used as a definition for test subject when it actually isn’t. We need to have a separate definition for the term “test subject” from how to refer to it in the output data. - Shadi

Edits for the current section:

- Need to spell out ATT. Abbreviation not explained
- Also non-ATT developers may need to refer to the state of an application.
[Accessibility Conformance Testing - Shadi and Anne comments.docx](https://github.com/w3c/wcag-act/files/2401357/Accessibility.Conformance.Testing.-.Shadi.and.Anne.comments.docx)

### LINKED ISSUE #272: Clear split between requirements for rule authors vs. rule implementers

### LINKED ISSUE #279: "Failed" / "Passed" or "Fail"/"Pass"?
Isn't "Failed" / "Passed" / "Inapplicable" the agreed upon terminology for outcomes of rules?
In general af Ctrl+F reveals a lot of "Fail" and "Pass" throughout the document.

### LINKED ISSUE #280: Mark all Informative sections as such
Mark all informative sections as informative for CR draft, e.g. Introduction.

### LINKED ISSUE #281: Scope sections states ACT rules should not be used for conformance
From Shadi, do we really mean to use should here? "ACT Rules should not be used for conformance claims unless the rule explicitly states it can be used that way. See Rule Aggregation for details."

WilcoFiers:
### Issues resolved by this PR: 

- Clear split between requirements for rule authors vs. rule implementers : Closes #272
- Need a definition of Test Target : Closes #271
- 16. Rule Aggregation - Add Requirement : Closes #252
- Accessibility Requirements: Updates requested (Shadi) : Closes #250
- 15.2. Accuracy Benchmarking - Update definition : Closes #236
- Outcomes of atomic tests : Closes #235
- clarifying the relationship between rule application and accessibility conformance : Closes #227
- ACT i18n checks : Closes #226
- Clarify the definition of Rule Aggregation : Closes #165
- Clarify how to do "Implementation validation" and "accuracy benchmarking" for manual or semi-automated rules : Closes #163
- Clarify the output format definition : Closes #162
- Mark all Informative sections as such : Closes #280
- "Failed" / "Passed" or "Fail"/"Pass"? : Closes #279
- Require Rule ID's in Atomic Rule List : Closes #261
- Scope sections states ACT rules should not be used for conformance : Closes #281
- Section “13. Accessibility Support” unclear : Closes #221
- Examples need to be identified accessibly : Closes #187

### Outstanding discussions before CR: 

- List of features for exit criteria #224

### Outstanding editorials before CR:

- Need non-versioned biblio reference tags #216

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/pull/283.html" title="Last updated on Oct 14, 2018, 1:53 PM GMT (a34cafb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/283/38370ea...a34cafb.html" title="Last updated on Oct 14, 2018, 1:53 PM GMT (a34cafb)">Diff</a>


act-rules-format.bs
```diff
@@ -52,66 +48,91 @@ The separation between atomic rules and composite rules creates a division of re
...
 * [Applicability](#applicability)
 * [Expectations](#expectations)
-* [Limitations, Assumptions or Exceptions](#structure-limitations-assumptions-exceptions), if any
-* [Accessibility Support](#acc-support) (optional)
+* [Limitations, Assumptions or Exceptions](#limitations-assumptions-exceptions), if any
+* [Accessibility Support](#accessibility-support) (optional)
+* [Evaluation Instructions](#evaluation-instructions)
+* [Test Cases](#test-cases)
+
+ACT Rules MUST be written in an format that conforms to the Web Content Accessibility Guidelines [[WCAG]] or a comperable accessibility guidelines recommendation. ACT Rule [test cases](#test-cases) are allowed to contain inaccessible content. If any test case contains accessibility issues listed in [WCAG 2.1 Section 5.2.5 Non-Interference](https://www.w3.org/TR/WCAG21/#cc5), users MUST be warned of this in advance. Using an accessibility format supports people with disabilities. It also makes internationalisation of ACT rules easier.
```
maryjom: Editorial:

- First sentence: ...written in "a" format, not "an" format   and "comparable" instead of "comperable"
- First sentence: Suggest you say "comparable accessibility standard" since not all are called "guidelines" or are a "recommendation", like the EN 301 549 or other non-W3C standard.
- You should probably provide an explanation of why there MUST be a warning if the (do you really mean test case or do you mean rule?) contains accessibility issues listed in the Non-interference section of WCAG.

act-rules-format.bs
```diff
@@ -17,31 +17,27 @@ Introduction {#intro}
...
-
-Because ACT Rules rarely test the entirety of an accessibility requirement, passing ACT rules does not necessarily mean that an accessibility requirement is met. It is important to understand that **ACT Rules test non-conformance** to accessibility requirements. In some cases conformance can be inferred from the absence of failures. Unlike WCAG sufficient techniques, ACT Rules can not be used for conformance claims unless the rule explicitly states it can be used that way. See [Rule Aggregation](#rule-aggregation) for details.
+The ACT Rules Format can be used to describe ACT Rules dedicated to testing the [accessibility requirements](#accessibility-requirements) defined in Web Content Accessibility Guidelines [[WCAG]], which are specifically designed for web content. Other accessibility requirements applicable to web technologies can also be testable with ACT Rules. For example, ACT Rules could be developed to test the conformance of web-based user agents to the [User Agent Accessibility Guidelines](https://www.w3.org/WAI/standards-guidelines/uaag/) [[UAAG20]]. The ACT Rules Format may not always be suitable to describe tests for other types of accessibility requirements.
```
nitedog: "may" -> "might"

act-rules-format.bs
```diff
@@ -17,31 +17,27 @@ Introduction {#intro}
...
 
-The separation between atomic rules and composite rules creates a division of responsibility. Atomic rules test if web content is correctly implemented in a particular solution. Composite rules test if a combination of `pass` outcomes from atomic rules is sufficient for an accessibility requirement not to `fail`. Not all atomic rules have to be part of a composite rule. Atomic rules only have to be part of a composite rule if failing that atomic rule does not directly fail [Accessibility Requirements](#structure-accessibility-requirements). An atomic rule MAY be part of multiple composite rules.
+- <dfn>Composite rules</dfn> describe how the [=outcome=] of [=atomic rules=] should be used to decide if an accessibility requirement is (non-)conformant. If there are multiple `failed` outcomes necessary for an accessibility requirement to be (non-)conformant, each "type" of `failed` outcome would come from a separate atomic rules, and the logic for determining the conformance or non-conformance for the accessibility requirement, across the atomic rules, should be handled by a composite rule.
```
nitedog: "from a separate atomic rules" -> "from a separate atomic rule" (singular)

act-rules-format.bs
```diff
@@ -52,66 +48,91 @@ The separation between atomic rules and composite rules creates a division of re
     <li>Header indicated by using the `headers` attribute</li>
     <li>Header indicated by using ARIA labels</li>
   </ul>
-  <p>If any one of these rules passes, the test result of the composite rule for the table cell is "passed".</p>
+  <p>If any one of these rules passes, the outcome of the composite rule for the table cell is "passed".</p>
 </div>
 
+The separation between atomic rules and composite rules creates a division of responsibility. Atomic rules test if web content is correctly implemented in a particular solution. Composite rules test if a combination of outcomes from atomic rules is sufficient for an accessibility requirement to be conformant or not.
```
nitedog: "Atomic rules test if web content is correctly implemented in a particular solution" -> "Atomic rules test if test subjects correctly implement particular solutions" or "Atomic rules test for particular solutions within test subjects"

"Composite rules test if a combination of outcomes from atomic rules is sufficient for an accessibility requirement to be conformant or not." -> "Composite rules test if a combination of outcomes from atomic rules satisfy/meet the conformance for an accessibility requirement."

act-rules-format.bs
```diff
@@ -52,66 +48,91 @@ The separation between atomic rules and composite rules creates a division of re
...
+
+Not all atomic rules have to be part of a composite rule. Only if the outcome of multiple atomic rules has to be combined in order to determine (non-)conformance of the [Accessibility Requirements](#accessibility-requirements) can composite rules be used. An atomic rule MAY be part of multiple composite rules.
```
nitedog: "Only if the outcome of multiple atomic rules has to be combined in order to determine (non-)conformance of the [Accessibility Requirements](#accessibility-requirements) can composite rules be used" -> "Composite rules are used when the outcome of multiple atomic rules needs to be combined in order to determine (non-)conformance [of a test subject] to the [Accessibility Requirements](#accessibility-requirements)"

act-rules-format.bs
```diff
@@ -52,66 +48,91 @@ The separation between atomic rules and composite rules creates a division of re
...
+Accessibility Requirements {#accessibility-requirements}
+========================================================
 
-[=Atomic rules=] and [=composite rules=] SHOULD identify the accessibility requirements that fail when the outcome of a rule is Fail. An ACT Rule is a complete or partial test for one or more accessibility requirements. This means that most ACT rules test only part of an accessibility requirement, but it MUST NOT test more than the accessibility requirement it lists.
+Accessibility requirements are just that: A requirement that a particular web page must conform to for it to be considered accessible. A common example of accessibility requirements are the WCAG 2.1 success criteria. Often organizations have additional requirements which may come from different sources, such as laws, internal standards, etc. These too are considered accessibility requirements which can be tested using the ACT Rules Format.
```
nitedog: "web page" -> "test subject"
"must conform to" -> "conforms to" (drop the "must")
"Often organizations" -> "Some organizations"
"which may come from" -> "that come from" (drop the "may")

act-rules-format.bs
```diff
@@ -52,66 +48,91 @@ The separation between atomic rules and composite rules creates a division of re
...
+<div class=example>
+  <p>The outcome of this rule effects conformance to the following WCAG 2.1 success criteria:</p>
+  <ul>
+    <li>[1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships)</li>
+    <li>[4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value)</li>
+  </ul>
+</div>
+
+One reason why an [=atomic rule=] might not list accessibility requirements is because it is part of a [=composite rule=], and only a `failed` [=outcome=] in the composite rule would mean not satisfying the success criteria. Another reason might be that an ACT Rule is designed to evaluate a best practice, which may not be part of any accessibility standard.
```
nitedog: "may" -> "is" (drop the "may")

act-rules-format.bs
```diff
@@ -52,66 +48,91 @@ The separation between atomic rules and composite rules creates a division of re
...
+An aspect is a distinct part of the [=test subject=]. For example, rendering a particular piece of content to an end user involves multiple different technologies, some or all of which may be of use in an [atomic rule]. Some rules need to operate directly on the [Hypertext Transfer Protocol](https://tools.ietf.org/html/rfc7230) [[http11]] messages exchanged between a server and a client, while others need to operate on the [Document Object Model](https://dom.spec.whatwg.org) [[DOM]] tree exposed by a web browser. Rules may need to operate on several aspects simultaneously, such as both the HTTP messages and the DOM tree.
```
nitedog: "may be of use" -> "can be of use" (unless you really mean "MAY")
"may need to operate" -> "can operate"

act-rules-format.bs
```diff
@@ -195,15 +216,16 @@ Note that atomic rules in a composite rule MAY have different applicability. Bec
...
+
+The expectations describe what the requirements are for the [=test targets=] derived from the [applicability](#applicability).
 
 
 Expectations for Atomic Rules {#expectations-atomic}
 ----------------------------------------------------
 
-An [=atomic rule=] MUST contain one or more expectations. An expectation is an assertion that must be true about each [test target](#output-test-target) (see [Applicability](#applicability)). Each expectation MUST be distinct, unambiguous, and be written in plain language. Unlike in applicability, a certain level of subjectivity is allowed in expectations. Meaning that the expectation has only one possible meaning, but that meaning isn't fully quantifiable.
+An [=atomic rule=] MUST contain one or more expectations. An expectation is an assertion that must be true about each [=test target=] (see [Applicability](#applicability)). Each expectation MUST be distinct, unambiguous, and be written in plain language. Unlike in applicability, a certain level of subjectivity is allowed in expectations. Meaning that the expectation has only one possible meaning, but that meaning MAY not fully quantifiable.
```
nitedog: "that must be true" -> "that is true" (drop the "must")

act-rules-format.bs
```diff
@@ -195,15 +216,16 @@ Note that atomic rules in a composite rule MAY have different applicability. Bec
...
 
-When all expectations are true for a test target, the test target `passed` the rule. If one or more expectations are false, the test target `failed` the rule. If the atomic rule is used in a [=composed rule=], the composed rule may be `passed` when the atomic rule `failed`, depending on the [applicability](#applicability-composed) and [exxpectations-composed] of the composed rule.
+When all expectations are true for a test target, the test target `passed` the rule. If one or more expectations are false, the test target `failed` the rule. If the atomic rule is used in a composed rule, the outcome of the composed rule may be `passed` when the atomic rule `failed`, depending on the [expectations](#expectations-composed) of the composed rule.
```
nitedog: "composed" -> "composite" (twice)
"may be" -> "can be" (drop the "may)

Suggestion: "If the atomic rule is used in a composed rule, the outcome of the composed rule may be `passed` when the atomic rule `failed`" -> "When atomic rules are used in composite rules, the outcome of the composite rules can be `passed` when the outcome of individual atomic rule is `failed`"

act-rules-format.bs
```diff
@@ -235,158 +257,198 @@ When all expectations are true for a test target, the test target `passed` the r
...
+ACT Rules are designed to evaluate the conformance of content using web technologies to accessibility requirements. However, not every feature of a web technology is implemented in all assistive technologies or user agents that a website may need to support. The concept of [accessibility supported](https://www.w3.org/TR/WCAG20/#accessibility-supporteddef) use of a Web technology is defined in WCAG [[WCAG]]. Because of this, ACT Rules are not necessarily applicable in all test scenarios. For instance, a web page that has to work in assistive technologies that have no WAI-ARIA [[WAI-ARIA]] support, wouldn’t be tested with an ACT Rule that relies on WAI-ARIA support, since this rule could give results that are inconsistent with the accessibility experience of users of those assistive technologies.
```
nitedog: "that a website" -> "that content" (to be less website-centric)

act-rules-format.bs
```diff
@@ -235,158 +257,198 @@ When all expectations are true for a test target, the test target `passed` the r
...
 
 Even within a [=composite rule=], some [=atomic rules=] may not always be applicable. This leaves one fewer solution for passing that particular composite rule. To support users of ACT Rules in properly defining the [accessibility support baseline](https://www.w3.org/TR/WCAG-EM/#step1c) in their test scenarios, an ACT Rule SHOULD include a warning if there are significant accessibility support concerns known about a rule.
```
nitedog: "may not" -> "might not" (drop the "may")
"include a warning if there are" -> "indicate" (more concise; also not really a "warning", it is information like all other sections)

act-rules-format.bs
```diff
@@ -235,158 +257,198 @@ When all expectations are true for a test target, the test target `passed` the r
...
 
-ACT Data Format (Output Data) {#output}
-=======================================
-
-With ACT Rules, it is important that data coming from different sources can be compared. By having a shared vocabulary, accessibility data that is produced by different auditors can be compared and, where necessary, aggregated. Therefore, every ACT Rule MUST express the output in a format that has all of the features described in the ACT Data Format.
-
-Rules are tested in two steps. Firstly, the applicability is used to find a list of [Test Targets](#output-test-target) (elements, tags or other "components") within the web page or other [test subject](#output-test-subject). Then each test target is tested to see if all of the [expectations](#expectations) are true. This will give the outcome for each test target. For contextual information, the output data must also include [test subject](#output-test-subject) and the [rule identifier](#rule-identifier).
-
-This will mean that every time a rule is executed on a page, it will return a set with zero or more results, each of which MUST have at least the following properties:
+Evaluation Instructions {#evaluation-instructions}
```
nitedog: The name is a bit odd. I suggest "Accessibility Requirements Mapping" (see more below on why this name may be more apt)

act-rules-format.bs
```diff
@@ -235,158 +257,198 @@ When all expectations are true for a test target, the test target `passed` the r
...
+========================
 
-- [Rule Identifier](#rule-identifier) (test)
-- [Test Subject](#output-test-subject) (Web page)
-- [Test Target](#output-test-target) (pointer)
-- [Outcome](#output-outcome) (`Passed`, `Failed`, or `Inapplicable`)
+An ACT Rule MUST include evaluation instructions. Evaluation instructions describes how applicability and expecations can be used to determine (non-)conformance to the [listed accessibility requirements](#accessibility-requirements). Evaluation instructions MUST indicate how `passed`, `failed` and `inapplicable` [=outcomes=] can be determined for each [=test subject=], when evaluating the rule.
```
nitedog: The first sentence is already covered by section 4 "Rule Structure"; the second sentence is already covered by section 11 "Expectations"; and the third sentence is already covered by the definition of "outcomes". I think we can lose this entire paragraph. It re-defines existing normative requirements and definitions, which is very confusing.

act-rules-format.bs
```diff
@@ -235,158 +257,198 @@ When all expectations are true for a test target, the test target `passed` the r
...
-  }
-}
-```
+  <ol>
+    <li>If no parts of the test subject match the applicability, the outcome is `inapplicable`</li>
+    <li>Otherwise, if all expectations of the rule are true for each test target, the outcome is `passed`.</li>
+    <li>Otherwise the outcome is `failed`.<li>
 </div>
 
+If any [accessibility requirements](#accessibility-requirements) are listed, the ACT Rule MUST indicate how each outcome impacts the conformance to the listed accessibility requirements. For WCAG 2.1 Success Criteria, this means indicating whether or not the success criterion can be satisfied, or if further testing is needed.
```
nitedog: "impacts" -> "relates to"

Replace the second sentence with the following:
"Typically failing a rule means the associated accessibility requirement was not met. At the same time, passing rule does not usually mean that the associated accessibility requirement was met, unless the rule was designed to address all fail and pass conditions of that accessibility requirement."

act-rules-format.bs
```diff
@@ -235,158 +257,198 @@ When all expectations are true for a test target, the test target `passed` the r
...
-  "result": {
-    "outcome": "Failed",
-    "pointer": "html > body > h1:first-child"
-  }
-}
-```
+  <ol>
+    <li>If no parts of the test subject match the applicability, the outcome is `inapplicable`</li>
+    <li>Otherwise, if all expectations of the rule are true for each test target, the outcome is `passed`.</li>
+    <li>Otherwise the outcome is `failed`.<li>
```
nitedog: This example should also be dropped. This is rewriting the expectation, and seems unnecessary to me.

act-rules-format.bs
```diff
@@ -235,158 +257,198 @@ When all expectations are true for a test target, the test target `passed` the r
...
+  <ul>
+    <li>`failed`: The success criteria are *not satisfied*</li>
+    <li>`passed` or `inapplicable`: Further testing is necessary to determine if the success criteria are satisfied</li>
+  </ul>
+</div>
 
-<div class="note">
-  While *inapplicable* is a valid result for ACT Rules, it may not be a valid result for all [accessibility requirements](#structure-accessibility-requirements). Notably the success criteria of WCAG 2.0 and WCAG 2.1 can only be evaluated to true (passed) or false (failed). This translation of results is part of [rule aggregation](#rule-aggregation)
+<div class=note>
+  In WCAG 2.1, success criteria do not evaluate to `passed`, `failed` or `inapplicable`. Rather they can be *satisfied* (or not). (See the [WCAG 2.1 definition: satisfies a success criterion](https://www.w3.org/TR/WCAG21/#dfn-satisfies).) If a success criterion is *not satisfied* a web page can only conform if there is a conforming alternative version, as described in [WCAG 2.1 Conformance Requirement 1](https://www.w3.org/TR/WCAG21/#cc1).
```
nitedog: This is already described in the definition for outcome and can be removed from here.

act-rules-format.bs
```diff
@@ -235,158 +257,198 @@ When all expectations are true for a test target, the test target `passed` the r
...
 </div>
 
-<div class="note">
-  In addition to `Passed` `Failed` and `Inapplicable`, [[EARL10-Schema]] also defined an `Incomplete` outcome. While this should never be the outcome of a rule when applied in its entirety, it often happens that rules are only partially executed. For example, when applicability was automated, but the expectations have to be evaluated manually. Such "interim" results can be expressed with the "Incomplete" outcome.
+<div class=note>
+  In WCAG 2.1, all content must satisfy the success criteria. Because of this, all ACT Rules for WCAG 2.1 will require a `passed` outcome for every test target, in order for the test subject to be `passed`. While this logic of "every element must pass for the page to pass" will be common, it is not required by the ACT Rules Format.
```
nitedog: I don't understand what this means but I think it is already sufficiently addressed by my proposed rewrite, and by the example above.

act-rules-format.bs
WilcoFiers: Test subjects aren't required to conform to WCAG, web pages are. The way I see it, web pages are a type of test subject. JSX or Vue files are examples of test subjects that are not web pages. Instead, they are used to build a web page. WCAG does not apply to those files, nevertheless those can be tested by ACT Rules. I think this sentence is correct.

nitedog: I don't understand why you want to limit "accessibility requirements" to web page and WCAG only. As I understand it, accessibility requirements is more generic. It can be used for any requirement, for any test subject. We happen to only *currently* use it for WCAG and web pages in Auto-WCAG. Anyway, this sentence does not set a normative requirement and could be edited later, if we decide to.

WilcoFiers:
@nitedog @maryjom I've updated the PR. I think this is ready for survey.

act-rules-format.bs
```diff
@@ -52,66 +48,105 @@ The separation between atomic rules and composite rules creates a division of re
     <li>Header indicated by using the `headers` attribute</li>
     <li>Header indicated by using ARIA labels</li>
   </ul>
-  <p>If any one of these rules passes, the test result of the composite rule for the table cell is "passed".</p>
+  <p>If any one of these rules passes, the outcome of the composite rule for the table cell is "passed".</p>
 </div>
 
+The separation between atomic rules and composite rules creates a division of responsibility. Atomic rules test if web content is correctly implemented in a particular solution. Composite rules test if a combination of outcomes from atomic rules satisfy the conformance for an accessibility requirement.
```
rdeltour: This may be confusing as this separation of responsibility is not always the case, as stated in the following paragraph: sometimes an atomic rule itself would test for conformance to an entire accessibility requirement, sometimes it requires a composed rule. My suggestion is simply to drop this paragraph, or make it a "note" block with a slight reword like

> The separation (...) ~creates~ _can be seen as_ a division of responsibility. Atomic rules ~test~ _typically test_ if web content is correctly implemented in a particular solution, while composite rules ~test~ _would test_ if a combination of outcomes from atomic rules satisfy the conformance for an accessibility requirement ~.~ _as a whole_.”

act-rules-format.bs
```diff
@@ -170,12 +205,12 @@ An objective description is one that can be resolved without uncertainty in a gi
...
 
-A [=composite rule=] defines how the outcomes from its atomic rules are used to determine a single outcome for each applicable test target. 
+A [=composite rule=] defines how the [=outcomes=] from rules in its [atomic rules list](#atomic-rules-list) are used to determine a single outcome. 
 
-The applicability of a composite rule is defined as the union of all the applicability sections of its [=atomic rules=]. Because of this, applicability of a composite rule can be inferred from the atomic rules it uses results from. Since the applicability can be inferred, rule authors MAY omit applicability for a composite rule. This can be useful if it is difficult to express the combined applicability in plain language.
+The applicability of a composite rule MUST be the union of all the applicability sections of the rules in its [atomic rules list](#atomic-rules-list). Because of this, applicability of a composite rule can be inferred from the atomic rules it uses outcomes from. Since the applicability can be inferred, rule authors MAY omit applicability for a composite rule. This can be useful if it is difficult to express the combined applicability in plain language.
```
rdeltour: I preferred the previous version, saying “The applicability _is defined as_” rather than “The applicability _MUST be_”: if we say it's _defined_ as such, then there is no room for interpretation, and implementations don't have to guess what to do if the author didn't follow the spec and defined the applicability otherwise. If we use a MUST, then authors may (will) break this statement and implementations are left with an unspecified situation.

act-rules-format.bs
```diff
@@ -17,31 +17,27 @@ Introduction {#intro}
...
 
-The separation between atomic rules and composite rules creates a division of responsibility. Atomic rules test if web content is correctly implemented in a particular solution. Composite rules test if a combination of `pass` outcomes from atomic rules is sufficient for an accessibility requirement not to `fail`. Not all atomic rules have to be part of a composite rule. Atomic rules only have to be part of a composite rule if failing that atomic rule does not directly fail [Accessibility Requirements](#structure-accessibility-requirements). An atomic rule MAY be part of multiple composite rules.
+- <dfn>Composite rules</dfn> describe how the [=outcome=] of [=atomic rules=] is used to decide if an accessibility requirement is (non-)conformant. If this involves multiple "failure conditions", each of these is provided in a separate atomic rule. The logic for determining (non-)conformance to the accessibility requirement according to outcome of the individual the atomic rules is specified in the composite rule.
```
nitedog: - "decide if an accessibility requirement is (non-)conformant" does not make sense. Maybe "decide if a test subject conforms to an accessibility requirement"?
- Same with "logic for determining (non-)conformance to the accessibility requirement" -- needs "test subject" -- "logic for determining conformance of the test subject to the accessibility requirement"
- *BUT* if you define composite rules that way, then each composite rule *MUST* be related to an accessibility requirement. However, the requirement later on in section 7 is only a SHOULD requirement. You need to either change that requirement or this definition of composite rules.

act-rules-format.bs
```diff
@@ -52,66 +48,105 @@ The separation between atomic rules and composite rules creates a division of re
...
 
-Accessibility Requirements {#structure-accessibility-requirements}
-==================================================================
+Accessibility requirements are just that: A requirement that a particular web page conforms to for it to be considered accessible. A common example of accessibility requirements are the WCAG 2.1 success criteria. Some organizations have additional requirements that come from different sources, such as laws, internal standards, etc. These too are considered accessibility requirements which can be tested using the ACT Rules Format.
 
-<div class=note>
-  **Editor's note:** The ACT Task Force is looking for feedback about the use of the term "pass" in relation to rules. While rules can "pass", their corresponding accessibility requirement can fail. A section has been added to make this explicit in the rule, but we would like to know if this is sufficient.
+An ACT Rule SHOULD list an identifier and reference for each accessibility requirement that it maps to. Exactly how conformance to accessibility requirements is determined is described in the [Evaluation Instructions](#evaluation-instructions).
```
nitedog: - According to the definition of composite rules in section 3, they actually *MUST* identify the accessibility requirements they relate to (see corresponding comment above)
- There is no "Evaluation Instructions" section (anymore). I don't think that sentence is needed anyway.

act-rules-format.bs
```diff
@@ -52,66 +48,105 @@ The separation between atomic rules and composite rules creates a division of re
...
+One reason why an [=atomic rule=] might not list accessibility requirements is because it is part of a [=composite rule=], and only a `failed` [=outcome=] in the composite rule would mean not satisfying the accessibility requirement. Another reason might be that an ACT Rule is designed to evaluate a best practice that is not part of any accessibility standard.
 
-[=Atomic rules=] and [=composite rules=] SHOULD identify the accessibility requirements that fail when the outcome of a rule is Fail. An ACT Rule is a complete or partial test for one or more accessibility requirements. This means that most ACT rules test only part of an accessibility requirement, but it MUST NOT test more than the accessibility requirement it lists.
+If any accessibility requirements are listed, the ACT Rule MUST indicate how each outcome impacts the conformance to the listed accessibility requirements. For WCAG 2.1 Success Criteria, this means indicating whether or not the success criterion can be satisfied, or if further testing is needed.
```
nitedog: I suggest making that a sub-section 7.1 -- maybe call it "Rule Mapping" or such

act-rules-format.bs
tbostic32: Remove "the" between individual and atomic

act-rules-format.bs
kengdoj: agree with the reworded suggestion.

WilcoFiers: Discussed on the call. We'll leave it in place, not as a note but taking over the editorial changes

act-rules-format.bs
WilcoFiers: Need to remove "evaluation instructions".

act-rules-format.bs
WilcoFiers: Decision from the call is to leave the "MUST" in here, it is a requirement for the rule author. We can work on rewording this to address the issue separately.

nitedog: Suggestion: "The applicability of a composite rule is defined as the union of all the applicability sections of its [=atomic rules=]. This can be inferred from the list of atomic rules. However, rule authors MAY provide a description of the applicability for composite rules. This can be useful if it is difficult to express the combined applicability in plain language. In that case, the applicability of the composite rule provided by the author MUST also be the union of all the applicability sections of its."

act-rules-format.bs
```diff
@@ -37,7 +37,7 @@ In accessibility, there are often different technical solutions to make the same
...
 
-- <dfn>Composite rules</dfn> describe how the [=outcome=] of [=atomic rules=] is used to decide if an accessibility requirement is (non-)conformant. If this involves multiple "failure conditions", each of these is provided in a separate atomic rule. The logic for determining (non-)conformance to the accessibility requirement according to outcome of the individual the atomic rules is specified in the composite rule.
+- <dfn>Composite rules</dfn> describe how the [=outcome=] of multiple [=atomic rules=] can be combined into a single outcome. A composite rule can have multiple "satisfying conditions", each of these provided in a separate atomic rule. The logic in the composite rule describes how any one of these satisfying conditions, or some combination of them is used to determine a single outcome.
```
nitedog: - I really like the idea of saying "satisfying conditions" because it is not limited to "failure" or "success" but could be any. I think we need to reflect that phrasing in the atomic rules definition too. That is, "satisfying condition" instead of "failure condition".
- The last sentence seems to repeat the first two. Actually it contradicts them somehow -- is it "multiple" or "any one or more"? I'm not sure what the last sentence adds that is not already in the first two, and think it should be removed.
- Following the previous point, I think the word "multiple" is wrong in the first two sentences, regardless if you keep the third sentence or drop it. It should read "any one or more" instead of "multiple".

act-rules-format.bs
```diff
@@ -51,7 +51,7 @@ In accessibility, there are often different technical solutions to make the same
   <p>If any one of these rules passes, the outcome of the composite rule for the table cell is "passed".</p>
 </div>
 
-The separation between atomic rules and composite rules creates a division of responsibility. Atomic rules test if web content is correctly implemented in a particular solution. Composite rules test if a combination of outcomes from atomic rules satisfy the conformance for an accessibility requirement.
+The separation between atomic rules and composite rules can be seen as a division of responsibility. Atomic rules typically test if web content is correctly implemented in a particular solution. Composite rules would test if a combination of outcomes from atomic rules satisfy the accessibility requirement, in part or as a whole.
```
nitedog: I'm not the grammar expert but "would" jumps out at me. I think it should be "can".

act-rules-format.bs
nitedog: Looks good!

act-rules-format.bs
```diff
@@ -210,12 +219,12 @@ Applicability for Composite Rules {#applicability-composite}
 
 A [=composite rule=] defines how the [=outcomes=] from rules in its [atomic rules list](#atomic-rules-list) are used to determine a single outcome. 
 
-The applicability of a composite rule MUST be the union of all the applicability sections of the rules in its [atomic rules list](#atomic-rules-list). Because of this, applicability of a composite rule can be inferred from the atomic rules it uses outcomes from. Since the applicability can be inferred, rule authors MAY omit applicability for a composite rule. This can be useful if it is difficult to express the combined applicability in plain language.
+The applicability of a composite rule is defined as the union of all the applicability sections of its [=atomic rules=]. This can be inferred from the [list of atomic rules](#atomic-rules-list). Rule authors MAY describe the applicability for composite rules. This can be useful if it is difficult to express the combined applicability in plain language. If the composite rule includes applicbility, it MUST be the union of all the applicability sections of the atomic rules.
```
nitedog: applicbility -> applicability

act-rules-format.bs
maryjom: Suggest the last sentence reads, "The logic in the composite rule describes one or more of these satisfying conditions is used to determine a single outcome."

act-rules-format.bs
maryjom: I would simply delete the word 'would'.

act-rules-format.bs
```diff
@@ -83,7 +83,7 @@ Rule Identifier {#rule-identifier}
 An ACT Rule MUST have a unique identifier that can be any unique text, such as plain text, URL or a database identifier.
 
 <div class=example>
-  <p>ACT Rules in the Auto-WCAG community groups use file names as identifiers. They include a number for a WCAG 2.1 success criterion, and a handle that includes an element name or attribute:</p>
+  <p><strong>Example</strong>: ACT Rules in the Auto-WCAG community groups use file names as identifiers. They include a number for a WCAG 2.1 success criterion, and a handle that includes an element name or attribute:</p>
```
maryjom: Suggested edit: 'ACT Rules developed by the Auto-WCAG community groups use..."  

Also, to clarify understanding of the 2nd sentence...Do you mean, "They include the number corresponding to a WCAG 2.1 success criterion, combined with a handle that includes an element name or name being tested:"
Does that change characterize it more clearly without changing your meaning?

act-rules-format.bs
```diff
@@ -98,22 +98,27 @@ Rule Description {#rule-description}
...
-An ACT Rule SHOULD list an identifier and reference for each accessibility requirement that it maps to. Exactly how conformance to accessibility requirements is determined is described in the [Evaluation Instructions](#evaluation-instructions).
+
+Accessibility Requirements List {#accessibility-requirements-list}
+------------------------------------------------------------------
+
+An ACT Rule SHOULD list an identifier and reference for each accessibility requirement that it maps to.
 
 <div class=example>
-  <p>The outcome of this rule effects conformance to the following WCAG 2.1 success criteria:</p>
+  <p><strong>Example</strong>: The outcome of this rule effects conformance to the following WCAG 2.1 success criteria:</p>
```
maryjom: I think "affects" instead of "effects" is a better word.

act-rules-format.bs
```diff
@@ -242,7 +251,7 @@ An [=atomic rule=] MUST contain one or more expectations. An expectation is an a
 When all expectations are true for a test target, the test target `passed` the rule. If one or more expectations are false, the test target `failed` the rule. When atomic rules are used in composite rules, the outcome of the composite rules can be `passed` when the outcome of individual atomic rule is `failed`, depending on the [expectations](#expectations-composite) of the composite rule.
 
 <div class=example>
-  <p>A rule for labels of HTML `input` elements might have the following expectations:</p>
+  <p><strong>Example</strong>: A rule for labels of HTML `input` elements might have the following expectations:</p>
```
maryjom: Between this change and the previous change on line 227, I'm getting confused about the use of "applicability' and 'expectations' as the listed statements used as examples of each seem like the exact same kind of check.

act-rules-format.bs
```diff
@@ -350,7 +361,7 @@ While the ACT Rules Format is designed to stimulate harmonization, there are no
 Harmonization occurs when a group of rule implementors collectively accept the validity of an ACT Rule. For example, a community group of accessibility testing tool vendors could declare they've harmonized on a particular set of ACT Rules. Such a group can set acceptance criteria for rules, and have quality requirements that go beyond the ACT Rules Format. For example:
 
 <div class=example>
-  <p>The ACT EPUB group might have the following acceptance criteria:</p>
+  <p><strong>Example</strong>: The ACT EPUB group might have the following acceptance criteria:</p>
   <ul>
     <li>An ACT EPUB Rule is harmonized when it was approved members of 3 organisations, AND</li>
     <li>An ACT EPUB Rule is harmonized when it has 2 independent implementation</li>
```
maryjom: Last bullet should say, "implementations".

act-rules-format.bs
```diff
@@ -403,10 +416,10 @@ Definitions {#definitions}
       <li>**Failed**: The test subject does not meets all expectations</li>
```
maryjom: Use "meet" instead of "meets"

act-rules-format.bs
```diff
@@ -368,29 +379,31 @@ Definitions {#definitions}
...
+      <p><strong>Example:</strong></p>
       <ul>
         <li>An HTML page, including all embedded scripts, style and images<li>
         <li>An EPUB document</li>
         <li>A Vue component file</li>
       </ul>
     </div>
     <div class=note>
-      <p>Implementers using the [[EARL10-Schema]] can express the test subject on the [subject property](https://www.w3.org/TR/EARL10-Schema/#subject)</p>
+      <p><strong>Note</strong>: Implementers using the [[EARL10-Schema]] can express the test subject on the [subject property](https://www.w3.org/TR/EARL10-Schema/#subject)</p>
```
maryjom: I'm thinking it should say "**using** the [subject property]" rather than "on..."

act-rules-format.bs
```diff
@@ -368,29 +379,31 @@ Definitions {#definitions}
...
+      <p><strong>Example:</strong></p>
       <ul>
         <li>Nodes within an HTML page</li>
         <li>A period of time within a video</li>
         <li>A range of characters within a text document</li>
       </ul>
     </div>
     <div class=note>
-      <p>Implementers using the [[EARL10-Schema]] can express the test target on the [pointer property](https://www.w3.org/TR/EARL10-Schema/#pointer)</p>
+      <p><strong>Note</strong>: Implementers using the [[EARL10-Schema]] can express the test target on the [pointer property](https://www.w3.org/TR/EARL10-Schema/#pointer)</p>
```
maryjom: Same as previous comment, "using" instead of "on".

act-rules-format.bs
```diff
@@ -403,10 +416,10 @@ Definitions {#definitions}
...
     </div>
     <div class="note">
-      <p>Implementers using the [[EARL10-Schema]] can express the outcome on the [outcome property](https://www.w3.org/TR/EARL10-Schema/#outcome). In addition to `passed` `failed` and `inapplicable`, EARL 1.0 also defined an `incomplete` outcome. While this can not be the outcome of an ACT Rule when applied in its entirety, it often happens that rules are only partially evaluated. For example, when applicability was automated, but the expectations have to be evaluated manually. Such "interim" results can be expressed with the `incomplete` outcome.
+      <p><strong>Note</strong>: Implementers using the [[EARL10-Schema]] can express the outcome on the [outcome property](https://www.w3.org/TR/EARL10-Schema/#outcome). In addition to `passed` `failed` and `inapplicable`, EARL 1.0 also defined an `incomplete` outcome. While this can not be the outcome of an ACT Rule when applied in its entirety, it often happens that rules are only partially evaluated. For example, when applicability was automated, but the expectations have to be evaluated manually. Such "interim" results can be expressed with the `incomplete` outcome.
```
maryjom: Same comment, "using the [outcome property]" instead of "on".

Do we have 'incomplete' as a valid result defined in our list of outcomes?

act-rules-format.bs
WilcoFiers: Used "with" instead.

act-rules-format.bs
WilcoFiers: No, incomplete is not an outcome recognised in the ACT Rules Format.

Issue #290, opened 2018-10-18
Consider restricting requirement mapping


WilcoFiers:
Proposal was to restrict the accessibility requirement mapping to a finite list, such as:

- Passed: The requirement(s) are satisfied
- Passed: Further testing is necessary to determine if the success criteria are satisfied
- Inapplicable: The requirement(s) are satisfied
- Inapplicable: Further testing is necessary to determine if the success criteria are satisfied
- Failed: The requirement(s) are *not* satisfied
- Failed: Further testing is necessary to determine if the success criteria are satisfied

WilcoFiers:
I have two objections to restricting the requirement mapping in this way.

1. It forces rule authors to put aggregation logic into expectations. My concern with this is that expectations are written to get passed / failed outcomes for individual test targets, and not for aggregating the outcome for the page / test subject as a whole.

2. It opens back up the discussion of having composite rules use results of other composite rules. This idea was initially rejected because it would allow writing rules as a decision tree. While decision trees are certainly used in several accessibility tools, there are many tools that have a fairly flat structure and could not implement rules expressed in a decision tree. For that same reason the original procedural approach was also rejected, because tools using decision trees can't implement the procedural approach.

----

### Proposal

I do think flexibility in the accessibility requirements mapping is necessary to avoid the problems outlined above. But there is space between: "pick one of these" six, and "free form write how to aggregate". I would suggest that we add the following:

1. The accessibility requirements mapping MUST be objective. This ensures it can be automated.
2. Only the following information can be used to determine if the accessibility requirement was satisfied or not, or if further testing is needed:
	- The outcome of the test subject
	- The outcome of each test target
	- Whether or not content is embedded on the page using technologies other than those listed in test aspects.

annethyme:
Other suggestion: 

Accessibility requirements: 
- 3.3.2 # Labels and Instructions (Full test)
- 4.1.2 # Name, Role, Value (Partial test)

annethyme:
Possible mapping patterns, as Anne sees it: 
* “Input data” 
   * Rule in itself does not even map to a best practice (e.g. a rule that tests that `<img>` has `alt=""`, but this is only used as input in the rule that tests if images have an accessible name)
* Best practice test
   * Output cannot be mapped to satisfied / not satisfied for accessibility requirement
* Non-conformance test
   * Failed = not satisfied 
* Full test for the requirement 
   * Failed = not satisfied 
   * Passed = satisfied

[...]

Issue #282, opened 2018-10-07
Normative or Informative


goodwitch:
Dear ACT,

As you are building the data structure for rules, can you add an indicator that clearly defines if a rule is based on normative (required for compliance) or informative? 

This will be very useful for making sure that a person shooting for minimum compliance can clearly understand if they need to meet a rule (or if the rule is just one way to meet a requirement).

annethyme:
@goodwitch, I was just discussing with a colleague how to understand your comment here. 
As we read it, you suggest adding something like this: 

"Image has accessible name" rule
- normative for 1.1.1 Non-Text Content

"Skip to content link" rule
- informative for 2.4.1 Bypass Blocks

Is that correctly understood?

goodwitch:
Yes!  That is exactly what I meant!

WilcoFiers:
The group has taken a long hard look at this issue and we think we've resolved this. Rules now MUST indicate whether or not they are required for conformance to the accessibility requirements document which it was designed to test. Thank you for submitting the issue @goodwitch!

PR #295, opened 2018-10-27
Accessibility References

nitedog:
Added the concept of "related references", which is different from "accessibility requirement", to hopefully address issues #282 and #290 discussed during the f2f.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/pull/295.html" title="Last updated on Nov 6, 2018, 10:04 PM GMT (ec5897d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/295/b7a9d55...ec5897d.html" title="Last updated on Nov 6, 2018, 10:04 PM GMT (ec5897d)">Diff</a>

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -106,43 +106,64 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
 </div>
 
 
+Accessibility References {#accessibility-references}
+====================================================
+
+Accessibility references are conformance requirements, good practices, or other related provisions that the rule relates to. For example, this could be WCAG 2.1 success criteria when the rule is written to help determine conformance, or advisory techniques when the rule is written to help determine good practice implementation.
+
+Accessibility references can be [accessibility requirements](#accessibility-requirements) or [related references](#related-references). Rule authors MAY list any number of accessibility references that the rule has been written to test. Accessibility references are omitted when rules do not directly relate to particular accessibility references but can be used together with other rules as input, or when rules are designed to be part of a [=composite rule=]. This listing can be extended when rules are applied in other contexts.
+
+
 Accessibility Requirements {#accessibility-requirements}
-========================================================
+--------------------------------------------------------
 
-Accessibility requirements are just that: A requirement that a particular web page conforms to for it to be considered accessible. A common example of accessibility requirements are the WCAG 2.1 success criteria. Some organizations have additional requirements that come from different sources, such as laws, internal standards, etc. These too are considered accessibility requirements which can be tested using the ACT Rules Format.
+Accessibility requirements are normative requirements for conformance, such as WCAG 2.1 Success Criteria. Some organizations have additional requirements, such as laws and internal standards. These too are considered accessibility requirements.
 
-
-Accessibility Requirements List {#accessibility-requirements-list}
-------------------------------------------------------------------
-
-An ACT Rule SHOULD list an identifier and reference for each accessibility requirement that it maps to.
+An ACT Rule MUST list an identifier and reference for each accessibility requirement that it is written to directly test. That is, when the outcome of a rule determines meeting or not meeting the accessibility requirement it has been written for. This MAY be omitted when rules are designed to be part of a [=composite rule=].
 
 <div class=example>
-  <p><strong>Example</strong>: The outcome of this rule affects conformance to the following WCAG 2.1 success criteria:</p>
+  <p><strong>Example</strong>: The outcome of this rule indicates conformance to the following WCAG 2.1 success criteria:</p>
   <ul>
     <li>[1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships)</li>
     <li>[4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value)</li>
   </ul>
 </div>
 
-One reason why an [=atomic rule=] might not list accessibility requirements is because it is part of a [=composite rule=], and only a `failed` [=outcome=] in the composite rule would mean not satisfying the accessibility requirement. Another reason might be that an ACT Rule is designed to evaluate a best practice that is not part of any accessibility standard.
 
+Related References {#related-references}
+----------------------------------------
 
-Accessibility Requirements Mapping {#accessibility-requirements-mapping}
-------------------------------------------------------------------------
+Related references are non-normative accessibility practices, such as Advisory Techniques that enhance accessibility beyond minimum conformance. In some cases related references could imply conformance with a related accessibility requirement.
 
-If any accessibility requirements are listed, the ACT Rule MUST indicate how each outcome impacts the conformance to the listed accessibility requirements. For WCAG 2.1 Success Criteria, this means indicating whether or not the success criterion can be satisfied, or if further testing is needed.
+An ACT Rule MUST list an identifier and reference for each related reference that it is written to directly test. When related references are listed, the rule MUST indicate that these are not required for conformance. This MAY be omitted when rules are designed to be part of a [=composite rule=].
 
 <div class=example>
-  <p><strong>Example</strong>: The outcome of this rule maps to the accessibility requirements in the following manner:</p>
+  <p><strong>Example</strong>: The outcome of this rule indicates conformance beyond the related WCAG 2.1 success criteria:</p>
   <ul>
-    <li>`failed`: The success criteria are *not satisfied*</li>
-    <li>`passed` or `inapplicable`: Further testing is necessary to determine if the success criteria are satisfied</li>
+    <li>[G127: Identifying a Web page's relationship to a larger collection of Web pages](https://www.w3.org/TR/WCAG20-TECHS/G127.html)</li>
+  </ul>
+</div>
+
+
+Rule Mapping {#rule-mapping}
+----------------------------
+
+For all [accessibility requirements](#accessibility-requirements) and [related references](#related-references) listed, the ACT Rule MUST indicate how each outcome relates to the listed accessibility references. That is, the rule specifies if each accessibility reference is `satisfied`, `not satisfied`, or `undetermined` for the rule outcomes `passed`, `failed`, and `inapplicable`. This mapping MUST be consistent with the accessibility references.
+
+<div class=example>
+  <p><strong>Example</strong>: The outcome of this rule maps to the listed accessibility requirements in the following manner:</p>
+  <ul>
+    <li>`failed`: *not satisfied* (the success criteria are not met)</li>
+    <li>`passed` or `inapplicable`: *undetermined* (further testing is necessary to determine if the success criteria are satisfied)</li>
   </ul>
 </div>
 
 <div class=note>
   **Note**: In WCAG 2.1, success criteria do not evaluate to `passed`, `failed` or `inapplicable`. Rather they can be *satisfied* (or not). (See the [WCAG 2.1 definition: satisfies a success criterion](https://www.w3.org/TR/WCAG21/#dfn-satisfies).) If a success criterion is *not satisfied* a web page can only conform if there is a conforming alternative version, as described in [WCAG 2.1 Conformance Requirement 1](https://www.w3.org/TR/WCAG21/#cc1).
+</div>
+
+<div class=note>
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
 </div>
 
-One reason why an [=atomic rule=] might not list accessibility requirements is because it is part of a [=composite rule=], and only a `failed` [=outcome=] in the composite rule would mean not satisfying the accessibility requirement. Another reason might be that an ACT Rule is designed to evaluate a best practice that is not part of any accessibility standard.
 
+Related References {#related-references}
+----------------------------------------
 
-Accessibility Requirements Mapping {#accessibility-requirements-mapping}
-------------------------------------------------------------------------
+Related references are non-normative accessibility practices, such as Advisory Techniques that enhance accessibility beyond minimum conformance. In some cases related references could imply conformance with a related accessibility requirement.
```
WilcoFiers: > In some cases related references could imply conformance with a related accessibility requirement

WHat does it mean to "imply conformance"? Do you mean something like not skipping heading levels is an indication that 1.3.1 was satisfied? I'm not sure this paragraph explains the concept of related references very well.

act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
+  <p><strong>Example</strong>: The outcome of this rule indicates conformance to the following WCAG 2.1 success criteria:</p>
   <ul>
     <li>[1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships)</li>
     <li>[4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value)</li>
   </ul>
 </div>
 
-One reason why an [=atomic rule=] might not list accessibility requirements is because it is part of a [=composite rule=], and only a `failed` [=outcome=] in the composite rule would mean not satisfying the accessibility requirement. Another reason might be that an ACT Rule is designed to evaluate a best practice that is not part of any accessibility standard.
 
+Related References {#related-references}
```
WilcoFiers: I don't think "references" is the right term here. I'm having a hard time coming up with an alternative, as I'm not entirely sure what this section is even for.

act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
+----------------------------------------
 
-Accessibility Requirements Mapping {#accessibility-requirements-mapping}
-------------------------------------------------------------------------
+Related references are non-normative accessibility practices, such as Advisory Techniques that enhance accessibility beyond minimum conformance. In some cases related references could imply conformance with a related accessibility requirement.
 
-If any accessibility requirements are listed, the ACT Rule MUST indicate how each outcome impacts the conformance to the listed accessibility requirements. For WCAG 2.1 Success Criteria, this means indicating whether or not the success criterion can be satisfied, or if further testing is needed.
+An ACT Rule MUST list an identifier and reference for each related reference that it is written to directly test. When related references are listed, the rule MUST indicate that these are not required for conformance. This MAY be omitted when rules are designed to be part of a [=composite rule=].
```
WilcoFiers: > related reference that it is written to directly test

I'm not too sure what it means to directly test a thing, if that thing only "implies conformance".

> This MAY be omitted when rules are designed to be part of a [=composite rule=].

"this" is ambiguous here. I take it you mean that related references may be omitted?

act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
 
 <div class=example>
-  <p><strong>Example</strong>: The outcome of this rule maps to the accessibility requirements in the following manner:</p>
+  <p><strong>Example</strong>: The outcome of this rule indicates conformance beyond the related WCAG 2.1 success criteria:</p>
```
WilcoFiers: I don't understand what "indicates conformance beyond" means.

act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
-    <li>`passed` or `inapplicable`: Further testing is necessary to determine if the success criteria are satisfied</li>
+    <li>[G127: Identifying a Web page's relationship to a larger collection of Web pages](https://www.w3.org/TR/WCAG20-TECHS/G127.html)</li>
+  </ul>
+</div>
+
+
+Rule Mapping {#rule-mapping}
+----------------------------
+
+For all [accessibility requirements](#accessibility-requirements) and [related references](#related-references) listed, the ACT Rule MUST indicate how each outcome relates to the listed accessibility references. That is, the rule specifies if each accessibility reference is `satisfied`, `not satisfied`, or `undetermined` for the rule outcomes `passed`, `failed`, and `inapplicable`.
```
WilcoFiers: I don't agree with introducing a new "undetermined" term here. "undetermined" is not a valid outcome for WCAG.

> relates to
This paragraph needs to define "mapping", so that it can be used in the next paragraph.

act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
+
+Every mapping MUST be consistent with the related accessibility references. When rules refer to accessibility requirements, the mapping MUST result in either `satisfied` or `not satisfied` for any of the rule outcomes. That is, the mapping is not `undetermined` for all outcomes when mapping to accessibility requirements. Such mappings indicate related references rather than requirements.
```
WilcoFiers: So `failed` could mean the SC is undetermined, as long as `passed` means the SC is satisfied? I'm not sure about the implications of this. That would let you write rules like "don't use iframe, video or object elements in HTML" to pass 1.2.3. I'm not sure I fully understand the implications of allowing this.

act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
+    <li>`passed` or `inapplicable`: *undetermined* (further testing is necessary to determine if the success criteria are satisfied)</li>
   </ul>
 </div>
 
 <div class=note>
   **Note**: In WCAG 2.1, success criteria do not evaluate to `passed`, `failed` or `inapplicable`. Rather they can be *satisfied* (or not). (See the [WCAG 2.1 definition: satisfies a success criterion](https://www.w3.org/TR/WCAG21/#dfn-satisfies).) If a success criterion is *not satisfied* a web page can only conform if there is a conforming alternative version, as described in [WCAG 2.1 Conformance Requirement 1](https://www.w3.org/TR/WCAG21/#cc1).
 </div>
 
+<div class=note>
+  **Note**: ACT Rules are commonly partial tests for accessibility references. That is, the outcomes `passed`, `failed`, and `inapplicable` map to `undetermined`. A rule is a complete test when the outcomes only map to `satisfied` and `not satisfied`.
```
WilcoFiers: That would mean in WCAG you can't have rules that are a "complete test". A HTML page title rule is inapplicable for PDF. Perhaps if passed and failed map to satisfied and not satisfied? I'm not sure that works.

act-rules-format.bs
nitedog: I addressed your other comments on this section. Hopefully it is now clearer.

act-rules-format.bs
nitedog: It is another word for what you previously called "needs further testing". IMHO it is nothing new and factually occurs. For example, when a rule is not applicable the implication on the SC is usually "undetermined" or "needs further testing" (choose your favorite label here).

act-rules-format.bs
nitedog: I reworded that to be clearer. Again, I think nothing is new here. We just want to avoid having an accessibility requirement where all outcomes imply "further testing is needed". I think we agreed on that during the f2f.

WilcoFiers:
My objection is two-fold. Firstly I disagree with associating non-requirements with WCAG success criteria. There are loads of examples that show that people don't properly understand it if you do that. The ACT Rules Format doesn't prevent anyone from adding references to WCAG into their best practices if that is the approach they prefer. But that is only one way to deal with best practices. BBC has its own standard instead, that's a second way to add best practices. Deque uses taxonomy and has a "best practices" category, that's a third. There are multiple ways to handle best practices, and in my opinion, the ACT Rules Format should not express just one of them.

Secondly, this proposal knocks out a load-bearing wall of the ACT idea. Rules test failures of accessibility requirements. That's the premise of this work, it's also one of the requirements we wrote into the ACT Framework Requirements. Accessibility requirements can be WCAG criteria, it can be the RGAA, it can be company policy, that part does not matter (also in the requirements for ACT). What matter, is that failing the rule, means not satisfying an accessibility requirement. We can't drop one of our requirements 5 minutes before publishing. If we really want that (which I question) than we need to sit down and reevaluate our position, and not try to rush into CR.

act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
 
 
-Accessibility Requirements {#accessibility-requirements}
-========================================================
+Accessibility References {#accessibility-references}
+====================================================
+
+Accessibility references are conformance requirements, good practices, or other related provisions that the rule relates to. For example, this could be WCAG 2.1 success criteria when the rule is written to help determine conformance, or advisory techniques when the rule is written to help determine good practice implementation.
+
+Accessibility references can be [accessibility requirements](#accessibility-requirements) or [related references](#related-references). Rule authors MAY list any number of accessibility references that the rule has been written to test. Accessibility references are omitted when rules do not directly relate to particular accessibility references but can be used together with other rules as input, or when rules are designed to be part of a [=composite rule=]. This listing can be extended when rules are applied in other contexts.
```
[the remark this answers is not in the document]
nitedog: Refined wording and added example to address these two comments.

act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
 
-Accessibility requirements are just that: A requirement that a particular web page conforms to for it to be considered accessible. A common example of accessibility requirements are the WCAG 2.1 success criteria. Some organizations have additional requirements that come from different sources, such as laws, internal standards, etc. These too are considered accessibility requirements which can be tested using the ACT Rules Format.
 
+Accessibility Requirements {#accessibility-requirements}
+--------------------------------------------------------
 
-Accessibility Requirements List {#accessibility-requirements-list}
-------------------------------------------------------------------
+Accessibility requirements are normative requirements for conformance, such as WCAG 2.1 Success Criteria. Some organizations have additional requirements, such as laws and internal standards. These too are considered accessibility requirements.
```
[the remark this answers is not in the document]
nitedog: Removed use of "normative" to avoid such misunderstandings.

act-rules-format.bs
```diff
@@ -106,45 +106,68 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
...
 
-An ACT Rule SHOULD list an identifier and reference for each accessibility requirement that it maps to.
+An ACT Rule MUST list an identifier and reference for each accessibility requirement that it is written to directly test. That is, when the outcome of a rule determines meeting or not meeting the accessibility requirement it has been written for. This MAY be omitted when rules are designed to be part of a [=composite rule=].
```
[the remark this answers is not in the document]
nitedog: Refined wording to address comments. As to omitting references/identifiers, I think this is taken from the previous wording. For example, when it is an atomic rule. Suggestions welcome.

act-rules-format.bs
nitedog: Removed sentence.

act-rules-format.bs
nitedog: Refined wording to address comments.

act-rules-format.bs
nitedog: Refined wording to address comments.

act-rules-format.bs
nitedog: If the rule only checks HTML, then it is by definition not a complete test for WCAG success criteria, which are technology-agnostic.

cpandhi:
Would qualifying the accessibility requirement in the ACT rule help? 
For example: Accessibility requirement:
WCAG 2.0 SC xxx
BBC internal requirement
IBM internal requirement

Issue #293, opened 2018-10-25
Mapping "Background" and "Assumptions" sections from Auto-WCAG to ACT Rules Format


annethyme:
Auto-WCAG has a "Background" section. Should this be reflected in the ACT Rules Format? 

"Limitations, Assumptions or Exceptions" in the ACT Rules Format is always called "Assumptions" in Auto-WCAG. 

Should either Auto-WCAG or ACT Rules Format align?

annethyme:
I apparently raised the same issue in the Auto-WCAG CG: https://github.com/auto-wcag/auto-wcag/issues/350

annethyme:
Decision in ACT call on November 15: 

- @skotkjerra writes up draft for first part.
- "Limitations, Assumptions or Exceptions" should be changed into "Assumptions"

skotkjerra:
As agreed in yesterdays meeting the linked PR is a proposal for resolving the issue of optional information and renaming the assumptions as a result of cooperative efforts between @annethyme and myself

moekraft:
Addressed by PR #305

PR #305, opened 2018-11-16
"Optional items" section + renaming "Assumptions"

annethyme:
Based on discussion of https://github.com/w3c/wcag-act/issues/293

***
<a href="https://pr-preview.s3.amazonaws.com/annethyme/wcag-act/pull/305.html" title="Last updated on Nov 29, 2018, 11:03 PM GMT (4bf90aa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/305/809e9f5...annethyme:4bf90aa.html" title="Last updated on Nov 29, 2018, 11:03 PM GMT (4bf90aa)">Diff</a>

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -73,7 +73,7 @@ An ACT Rule MUST consist of at least the following items:
 * [Accessibility Support](#accessibility-support)
 * [Test Cases](#test-cases)
 * [Change Log](#change-log)
-* [Issues List](#issues-list) (optional)
+* [Issues List](#issues-list)
 * [Acknowledgements](#acknowledgements) (optional)
 * [Background](#background) (optional)
 * ACT Rules Format [=outcome=] definition
@@ -130,8 +130,6 @@ An ACT Rule SHOULD list an identifier and reference for each accessibility requi
 </div>
 
 One reason why an [=atomic rule=] might not list accessibility requirements is because it is part of a [=composite rule=], and only a `failed` [=outcome=] in the composite rule would mean not satisfying the accessibility requirement. Another reason might be that an ACT Rule is designed to evaluate a best practice that is not part of any accessibility standard.
-
-This section MAY be left blank.
 
 Accessibility Requirements Mapping {#accessibility-requirements-mapping}
 ------------------------------------------------------------------------
@@ -292,7 +290,7 @@ An ACT Rule MUST list any assumptions, limitations or any exceptions for the eva
 
 Sometimes there are multiple plausible ways that an accessibility requirement can be interpreted. For instance, it is not immediately obvious if emoji characters are "text" or "non-text content" under WCAG 2.1. Whatever the interpretation is, this MUST be documented in the rule.
 
-This section MAY be left blank.
+While this item MUST be included in the ACT Rule, it MAY be empty when there are no known assumptions, limitations or exceptions.
 
 Accessibility Support {#accessibility-support}
 ==============================================
@@ -300,7 +298,7 @@ Content can be designed to rely on the support for particular accessibility feat
  
 An ACT Rule MUST include known limitations on accessibility support. For example, an atomic rule that checks for a particular accessibility feature that is known not to be widely supported by assistive technologies and web browsers, or a composite rule that includes atomic rules with known accessibility support limitations.
 
-This section MAY be left blank.
+While this item MUST be included in the ACT Rule, it MAY be empty.
  
 <div class=note>
   **Note:** WCAG Evaluation Methodology (WCAG-EM) provides guidance on defining an [accessibility support baseline](https://www.w3.org/TR/WCAG-EM/#step1c) for test scenarios, which can help users of ACT Rules to select the appropriate rules for their own circumstance.
@@ -325,7 +323,7 @@ Each new release of an ACT Rule MUST be identifiable with either a date or a ver
 </div>
 
 
-Issues List (optional) {#issues-list}
+Issues List {#issues-list}
 =====================================
 
 An ACT Rule MUST include a list or a reference to a list of any known issues. The issues list MUST be used to record cases where the [=outcome=] of an ACT Rule was `failed` where it ought to have been `passed` or `inapplicable`, or vice versa. There are several reasons why this might occur, including:
@@ -336,7 +334,7 @@ An ACT Rule MUST include a list or a reference to a list of any known issues. Th
 
 The issues list serves two purposes. For users of ACT Rules, the issues list gives insight into why an inaccurate result occurred, as well as provide confidence in the result of that rule. For the designer of the rule, the issues list is also useful to plan future updates to the rule. In a new version of the rule, resolved issues would be moved to the change log.
 
-This section can be left blank.
+While this item MUST be included in the ACT Rule document, it MAY be empty.
 
 Acknowledgements (optional) {#acknowledgements}
 ===============================================
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -328,7 +328,7 @@ An ACT Rule SHOULD include known limitations on accessibility support. For examp
 </div>
 
 
-Background {#Background}
+Background {#background}
 -----------------------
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -321,7 +313,38 @@ Each new release of an ACT Rule MUST be identifiable with either a date or a ver
...
+
+
+Accessibility Support {#accessibility-support}
+----------------------------------------------
+Content can be designed to rely on the support for particular accessibility features by different assistive technologies and user agents. For example, content using a particular WAI-ARIA [[WAI-ARIA]] feature relies on that feature to be supported by assistive technologies and user agents. This content would not work for assistive technologies and user agents that do not support WAI-ARIA. WCAG [[WCAG]] provides a definition for [accessibility supported](https://www.w3.org/TR/WCAG20/#accessibility-supporteddef) use of a Web technology.
+ 
+An ACT Rule SHOULD include known limitations on accessibility support. For example, an atomic rule that checks for a particular accessibility feature that is known not to be widely supported by assistive technologies and web browsers, or a composite rule that includes atomic rules with known accessibility support limitations.
```
WilcoFiers: Accessibility support falls under "should". I don't think it can be treated as optional.

annethyme: In the ACT Rules Structure section it said "Accessibility Support (optional)".
Do you disagree with that too?

Some of the others say ", if any" same place. Accessibility Support is the only one that had "(optional)" on it.

act-rules-format.bs
```diff
@@ -321,17 +325,33 @@ Each new release of an ACT Rule MUST be identifiable with either a date or a ver
...
 - The scenario did not exist (due to changing technologies) or was overlooked during the initial design of the rule.
 
 The issues list serves two purposes. For users of ACT Rules, the issues list gives insight into why an inaccurate result occurred, as well as provide confidence in the result of that rule. For the designer of the rule, the issues list is also useful to plan future updates to the rule. In a new version of the rule, resolved issues would be moved to the change log.
 
+This section can be left blank.
+
+Acknowledgements (optional) {#acknowledgements}
+===============================================
+
+An ACT Rule MAY contain acknowledgements. This may include, but is not limited to:
```
nitedog: Replace "This may include" with "This can include", to avoid using the keyword "may".

act-rules-format.bs
```diff
@@ -127,6 +131,7 @@ An ACT Rule SHOULD list an identifier and reference for each accessibility requi
 
 One reason why an [=atomic rule=] might not list accessibility requirements is because it is part of a [=composite rule=], and only a `failed` [=outcome=] in the composite rule would mean not satisfying the accessibility requirement. Another reason might be that an ACT Rule is designed to evaluate a best practice that is not part of any accessibility standard.
 
+This section MAY be left blank.
```
nitedog: I think this is redundant or I don't understand it. We have a SHOULD requirement already, so by definition it MAY be empty. Maybe you want to say that this heading MUST be in the rule document but that it MAY be empty?

act-rules-format.bs
```diff
@@ -280,21 +285,23 @@ When all expectations are true for a test target, the test target `passed` the r
...
+An ACT Rule MUST list any assumptions, limitations or any exceptions for the evaluation, the test environment, technologies being used or the subject being tested. For example, a rule that would partially test <a href="https://www.w3.org/WAI/WCAG20/quickref/#visual-audio-contrast-contrast" target="_blank">WCAG 2.0 Success Criterion 1.4.3 Contrast (Minimum)</a> based on the inspection of CSS properties could state that it is only applicable to HTML text content stylable with CSS, and that the rule does not support images of text.
 
 Sometimes there are multiple plausible ways that an accessibility requirement can be interpreted. For instance, it is not immediately obvious if emoji characters are "text" or "non-text content" under WCAG 2.1. Whatever the interpretation is, this MUST be documented in the rule.
 
+This section MAY be left blank.
```
nitedog: This is ambiguous. "An ACT Rule MUST list any assumptions ..." then "This section MAY be left blank" seem to be contradictory.

act-rules-format.bs
```diff
@@ -280,21 +285,23 @@ When all expectations are true for a test target, the test target `passed` the r
...
+ 
+An ACT Rule MUST include known limitations on accessibility support. For example, an atomic rule that checks for a particular accessibility feature that is known not to be widely supported by assistive technologies and web browsers, or a composite rule that includes atomic rules with known accessibility support limitations.
 
-An ACT Rule SHOULD include known limitations on accessibility support. For example, an atomic rule that checks for a particular accessibility feature that is known not to be widely supported by assistive technologies and web browsers, or a composite rule that includes atomic rules with known accessibility support limitations.
-
+This section MAY be left blank.
```
nitedog: Same as above: this is ambiguous. "An ACT Rule MUST include known limitations ..." then "This section MAY be left blank" seem to be contradictory.

act-rules-format.bs
```diff
@@ -321,17 +325,33 @@ Each new release of an ACT Rule MUST be identifiable with either a date or a ver
 </div>
 
 
-Issues List {#updates-issues}
--------------------------------------
+Issues List (optional) {#issues-list}
+=====================================
 
-An ACT Rule MAY include an issues list. When included, the issues list MUST be used to record cases when the [=outcome=] of an ACT Rule was `failed` where it ought to have been `passed`, or vice versa. There are several reasons why this might occur, including:
+An ACT Rule MUST include a list or a reference to a list of any known issues. The issues list MUST be used to record cases where the [=outcome=] of an ACT Rule was `failed` where it ought to have been `passed` or `inapplicable`, or vice versa. There are several reasons why this might occur, including:
```
nitedog: How can this section be marked as optional yet have a MUST requirement?

act-rules-format.bs
```diff
@@ -321,17 +325,33 @@ Each new release of an ACT Rule MUST be identifiable with either a date or a ver
...
 
 - Certain scenarios or the use of technologies that are very rare and were not included in the rule for that reason.
 - Certain accessibility features are impossible to test within the test environment. For instance, they might only be testable by accessing the accessibility API, require screen capturing, etc.
 - The scenario did not exist (due to changing technologies) or was overlooked during the initial design of the rule.
 
 The issues list serves two purposes. For users of ACT Rules, the issues list gives insight into why an inaccurate result occurred, as well as provide confidence in the result of that rule. For the designer of the rule, the issues list is also useful to plan future updates to the rule. In a new version of the rule, resolved issues would be moved to the change log.
 
+This section can be left blank.
```
nitedog: This entire section is optional, so by definition it can (MAY?) be left blank (same comment as above).

skotkjerra:
It says: MUST include a list or a reference to a list of any known issues. What if there are no known issues? Then it can/should be left blank.

skotkjerra:
What if there are no known limitations etc.? Then it should be left blank.

skotkjerra:
That is what we are trying to say to be more explicit on how to handle - leave section out or include heading with no content - i.e. leave blank.

nitedog:
My thoughts:

1. We do not have a concept of "headings" and "sections" in the spec. We just say "a rule must contain X". We would need to define "headings" and "sections" if we want to go down that route.
2. I think it is redundant and unnecessary. If there are "no known issues" then the MUST clause has been addressed. In my view the requirement already implies what needs to be included.
3. If you and others think we need to make this yet more explicit, then I suggest clarifying the requirement directly rather than having two separate requirements in two different areas of the section. For example, "A rule MUST include a list or a reference to a list of any known issues. This is [or 'MAY be' if you really want to make it a requirement] left out when there are no known issues."

annethyme:
@nitedog and @skotkjerra, would wording along these lines resolve any of the concerns? 

> "While this item MUST be included in the ACT Rule document, it MAY be empty."

nitedog:
Not my preference but I could live with it with some adjustments:

1. We need to re-state the conditions under which it is acceptable to leave the item empty (eg. "it MAY be empty when there are no known issues" or such).
2. The term "ACT Rule document" is not defined anywhere in the document. Either we define it and refer to it in section 4 (eg. "An ACT Rule document MUST consist of at least the following items:"), or we drop it from this statement (eg. "While this item MUST be included in the ACT Rule, it MAY be empty when there are no known issues").

act-rules-format.bs
annethyme: Good question! Definitely needed a cleanup.
Now it is MUST + not optional.
However, MAY + optional is also an option. 
Question is: if there are known issues, would it be okay not to list them in the ACT Rule?

annethyme:
Did as @nitedog suggested to make it less bad :) 
However, I am open to removing it again. We only put it in because it was brought up in the discussion in the ACT TF meeting.

annethyme:
@moekraft, 
In the ACT Rules Structure, 
 
* [Issues List](#issues-list) 

should be 

* [Issues List](#issues-list) (optional)

PR #307, opened 2018-11-23
Proposal of RDF schema for ACT

josehilera:
Hi, I think that could be be useful to provide RDF schemas or vocabularies for machine processing of ACT rules.

I propose a new folder named "schema", that includes the schemas, and a subfolder "examples" with:
- atomic and composed rules, coded as RDF files.
- examples or EARL reports
- examples or SPARQL queries to a possible RDF data set that included the schemas, rules, and EARL reports.

With the SPARQL query included as example in the archive sparql-query-exampe.ttl, what is intended is first to obtain the variable "?Pointer", which represents the position with CSS selector format of all the places on the web page in which there has been an accessibility evaluation with failure result, which is extracted of the report itself. Second, the variable "? Expectation" is generated, which must show the description of what the applied rule establishes that must be expected in order to pass the evaluation correctly. Finally, the variable "?GoodExampleCode" is used to obtain the examples of possible test cases that the rule applied in its property "autowcag:testCase" contains.

The result of the query must be: 
Pointer:  "html[lang]"
Expectation: "The test target has a non-empty lang or xml:lang attribute."@en
Good example: `"<html` `lang='en'>"`

Pointer: "html[lang]"
Expectation: "The test target has a non-empty lang or xml:lang attribute."@en
Good example: `"<html xml:lang='en'>"`
 
All the archives are coded with Turtle (ttl) format, but they can be automatically converted in JSON-LD format.

I think that the ACT recommendation could include as appendix, not only EARL examples coded in RDF (JSON-LD) as now happens in the Working Draft 1.0, but also examples of rules coded ion RDF (JSON-LD).

Best.


josehilera:
Sorry, the correct SPARQL result is:

Pointer: "html[lang]"
Expectation: "The test target has a non-empty lang or xml:lang attribute."@en
Good example: "<html lang='en'>"

Pointer: "html[lang]"
Expextation: "The test target has a non-empty lang or xml:lang attribute."@en
Good example: "<html xml:lang='en'>"

We explained recently this proposal in a congress [ATICA2018](http://www.cc.uah.es/Atica/Atica2018/) but the paper is in Spanish. The examples are in english. The paper is attached.

[atica2018_Hilera.pdf](https://github.com/w3c/wcag-act/files/2611542/atica2018_Hilera.pdf)

WilcoFiers:
Hi Jose. Thank you for sharing these ideas. Can you explain to me what the use cases for this are? Who needs ACT Rules to be expressed in RDF, and what do they need it for?

josehilera:
The users of the vocabulary or scheme of ACT in RDF would be the following:

- **Authors of rules**: to express rules using the RDF vocabulary and be able to link the different fields of a rule with elements of other existing RDF vocabularies. For example, [schema.org](https://schema.org/), [foaf](http://xmlns.com/foaf/spec/), [dbpedia ](https://dbpedia.org/sparql) for the field "author". Also in the future, specific vocabularies could be created for fields such as limitations, assumpitons, agregationApplicability, etc., which allow the contents of these fields to be semantically formalized so that they can be processed, or even make inferences; something that is not possible if their content is expressed with a free text. For example, the accessibilityRequirement field can already be linked to existing RDF vocabularies for WCAG, such as those created by the [ACCESSIBLE ](http://www.accessible-eu.org/index.php/ontology.html)project. But it can also be linked to other accessibility specifications other than WCAG, such as the European [EN 501 349](http://www.etsi.org/deliver/etsi_en/301500_301599/301549/01.01.02_60/en_301549v010102p.pdf). In addition when using a formal language, the authors of rules can make automatic validations of the rules they are creating.
- **Authors of accessibility evaluation tools**: To be able to semantically enrich the EARL reports generated by their tool. The idea is to link the reports with the rules that have been applied in them, and this can be done if the report is expressed in the existing [RDF vocabulary for EARL](https://www.w3.org/TR/EARL10-Schema/), and is linked to the new RDF vocabulary for ACT. 
- **Accessibility evaluators**: The main utility would be for people who evaluate the accessibility of a website, because if they have a report expressed in the vocabulary RDF of EARL, and linked to rules expressed with the vocabulary RDF of ACT, this will open great possibilities, for example to compose more complex reports, to be able to search in the reports, that extend to the RDF vocabularies linked to the report through the rules.
- **Rules integrators**: This would be a new figure, which would be an accessibility expert knowledgeable of rules that have been created by different groups, not only by [Auto-WCAG](https://auto-wcag.github.io/auto-wcag/), but by different ones, and that generate knowledge bases in RDF, in which can express the equivalences between rules, or even define composite rules, in which the atomic rules come from different sources, for example adding an atomic rule created by Auto-WCAG and another created in another context or different expert group, or following the legislation of a particular country. These knowledge bases could also include rules of inference expressed in languages such as [SWRL](https://www.w3.org/Submission/SWRL/).

The vocabulary RDF for ACT would be the first step for, given an EARL report on accessibility of a website, anyone can make queries about it as the following:
- What rules have been applied in sections where the accessibility test has not been passed?
- Who was the author of those rules?
- What expectations were expected in an accessibility requirement that has not been passed?
- What is the description of the compliance criteria of the requirements that have been passed?
- In this evaluation report have rules been applied that guarantee that the accessibility requirements demanded by the legislation of country X are satisfied?
- If two reports have been obtained with two different evaluation tools on the same web page, have the same rules been applied by both tools, and if not, which rules prevail?
- Etc…

In short, have an RDF vocabulary about ACT allows to integrate into the universe of [Linked Data](https://www.w3.org/standards/semanticweb/data), supported by W3C, and to follow the path opened by other W3C recommendations on which W3C itself or other authors have published vocabularies or RDF schemes , such as: [WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#a_impl_roles), [HTTP](https://www.w3.org/TR/HTTP-in-RDF/), [HTML, WCAG2, CSS](http://www.accessible-eu.org/index.php/ontology.html), etc.

WilcoFiers:
Hey Jose. Thank you for the response. We've put this topic on the agenda of the ACT Task Force meeting this week.

I quite like the idea of having a format like that, but my personal view on it is that it is probably a bit too early for us to look into something like this. In my view, we need to focus on publishing rules first, streamline that process. Once we have that, I think it's worth seeing if enough organisations are interested in taking the work to that level. Again, that's my personal view, I'm looking forward to hear what the rest of the group thinks.

Issue #292, opened 2018-10-25
Create introduction to "What is an ACT rule?" in the spec


annethyme:
Per break-out session at TPAC on [Spec Editing Best Practices](https://lists.w3.org/Archives/Public/spec-prod/2018OctDec/0011.html), Wilco and Anne realised that there is currently nothing in the ACT Rules Format that introduces briefly/overall to "What is an ACT Rule" before we dive into the specifics of rule types and structure. 

We should have an introduction like this in the ACT Rules Format.

maryjom:
See pull request #310

maryjom:
This is done and approved.

PR #310, opened 2018-11-29
What is an ACT rule - Issue 292

maryjom:
See Issue #292


act-rules-format.html
```diff
@@ -342,6 +342,7 @@ <h2 class="heading settled" data-level="1" id="intro"><span class="secno">1. </s
...
 		<p>Describing how to test certain accessibility requirements will result in accessibility tests that are transparent, with test results that are reproducible. The Accessibility Conformance Testing (ACT) Rules Format defines the requirements for these test descriptions, known as Accessibility Conformance Testing Rules (ACT Rules).</p>
+		<p>Each ACT Rule is a plain English description of what type of content is being tested, the test it performs, and the expected result.  Where the test result affects conformance, the rule documents which particular requirement the test is for. The resulting outcomes from the test can be used to help determine conformance or non-conformance to the requirement. Test cases are also written as part of the ACT rule to provide a way to verify that implementations of the rule can successfully determine the expected pass and fail outcomes.</p>
```
annethyme: "passed" and "failed" outcomes?

annethyme: Does ACT Rules have to be in English? Can we say "plain language description" and allow for ACT Rules written in other languages?

annethyme: Should it say something about that ACT Rules Format defines requirements for which items a rules must contain to be an ACT Rule, or that ACT Rules have a certain structure?
And maybe also that there are different types of ACT Rules?

It would be nice if this intro was a high level summary of "everything in the spec".

WilcoFiers: This change should be made to the markdown file, not the HTML file. Otherwise we'd lose it.

moekraft: I think we should avoid passive English as well, e.g. "being tested" should rather be "content under test".

moekraft: Note: Updates are to HTML file and need to be made against Bikeshed file, act-rules-format.bs.

moekraft:
@maryjom Closing pull request to get changes in .bs file. Thanks, Moe

Issue #284, opened 2018-10-18
Proposal to (slightly) rearange the spec outline


rdeltour:
TL;DR: Can we consider grouping back the sections describing ACT Rules content together? 

I know we've been back and forth about this, and it's quite late in the process, but as I review the draft proposed to be sent to CR I find the current flat outline to be very slightly difficult to navigate.

I'm proposing:
- moving section 5 to 13 (from "Rule Identifier" to "Accessibility Support") under section 4 ("ACT Rule Structure")
- renaming "ACT Rule Structure" to "ACT Rule Content"

moekraft:
@rdeltour Hi Romain, I just want to confirm the updates requested. Are you asking that Rule Identifier move it's location to after Accessibility Support? 
Descriptive Title
~~Rule Identifier~~
Rule Description
Rule type
Accessibility Requirements
Aspects Under Test (for atomic rules) OR Atomic Rules List (for composite rules)
Applicability
Expectations
Assumptions
Accessibility Support
**Rule Identifier** <==Move here?
Test Cases
Change Log
Issues List
Acknowledgements (optional)
Background (optional)
ACT Rules Format outcome definition

rdeltour:
No, I was meaning to the rule-structure sections one level down:

1. Introduction
1. Scope
1. ACT Rule Types
1. ACT Rule Structure
    1. Rule Identifier
    1. Rule Description
    1. Accessibility Requirements
    1. Aspects Under Test (Atomic rules only)
    1. Atomic Rules List (Composite rules only)
    1. Applicability
    1. Expectations
    1. Assumptions
    1. Accessibility Support
    1. Test Cases
    1. Change Log
    1. Issues List (optional)
    1. Acknowledgements (optional)
    1. Background (optional)
1. Rule Accuracy
1. Harmonization
1. Definitions

instead of the current flat structure:

1. Introduction
1. Scope
1. ACT Rule Types
1. ACT Rule Structure
1. Rule Identifier
1. Rule Description
1. Accessibility Requirements
1. Aspects Under Test (Atomic rules only)
1. Atomic Rules List (Composite rules only)
1. Applicability
1. Expectations
1. Assumptions
1. Accessibility Support
1. Test Cases
1. Change Log
1. Issues List (optional)
1. Acknowledgements (optional)
1. Background (optional)
1. Rule Accuracy
1. Harmonization
1. Definitions

moekraft:
Got it. And also renaming to Rule Content

[...]

PR #322, opened 2019-01-15
Updates for CR

WilcoFiers:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/pull/322.html" title="Last updated on Jan 24, 2019, 9:05 PM UTC (a290814)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/322/79a05b3...a290814.html" title="Last updated on Jan 24, 2019, 9:05 PM UTC (a290814)">Diff</a>

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -15,9 +15,9 @@ Markup Shorthands: markdown yes
 Introduction {#intro}
 =====================
 
-There are currently many test procedures and tools available which aid their users in testing web content for conformance to accessibility standards such as the [Web Content Accessibility Guidelines (WCAG)](https://www.w3.org/WAI/standards-guidelines/wcag/) [[WCAG]]. As the web develops in both size and complexity, these procedures and tools are essential for managing the accessibility of resources available on the web.
+There are currently many test procedures and tools available which aid their users in testing web content for conformance to accessibility standards such as the [Web Content Accessibility Guidelines](https://www.w3.org/WAI/standards-guidelines/wcag/) [[WCAG]]. As the web develops in both size and complexity, these procedures and tools are essential for managing the accessibility of resources available on the web.
 
-This format is intended to enable a consistent interpretation of how to test conformance to [=accessibility requirements documents=] and promote consistent results in accessibility testing. The rules format is designed to describe both manual accessibility tests as well as automated tests as performed by accessibility testing tools.
+This format is intended to enable a consistent interpretation of how to test conformance to WCAG and other [=accessibility requirements documents=] and promote consistent results in accessibility testing. The rules format is designed to describe both manual accessibility tests, as well as automated tests as performed by accessibility testing tools.
 
 Documenting how to test [=accessibility requirements=] will result in accessibility tests that are transparent, with test results that are reproducible. The Accessibility Conformance Testing (ACT) Rules Format defines the requirements for these test descriptions, known as Accessibility Conformance Testing Rules (ACT Rules).
 
@@ -25,7 +25,7 @@ Documenting how to test [=accessibility requirements=] will result in accessibil
 Scope {#scope}
 ==============
 
-The ACT Rules Format defined in this specification is designed for rules that can be used in testing content created using web technologies, such as [Hypertext Markup Language](https://www.w3.org/TR/html/) [[HTML]], [Cascading Style Sheets](https://www.w3.org/TR/CSS/) [[css-2018]], [Accessible Rich Internet Applications](https://www.w3.org/WAI/standards-guidelines/aria/) [[WAI-ARIA]], [Scaleable Vector Graphics](https://www.w3.org/TR/SVG/) [[SVG2]], [EPUB](http://www.idpf.org/epub/31/spec/epub-spec.html), and more. The ACT Rules Format is designed to be technology agnostic, meaning that it can conceivably be used to describe test rules for other technologies.
+The ACT Rules Format defined in this specification is designed for rules that can be used in testing content created using web technologies, such as [Hypertext Markup Language](https://www.w3.org/TR/html/) [[HTML]], [Cascading Style Sheets](https://www.w3.org/TR/CSS/) [[css-2018]], [Accessible Rich Internet Applications](https://www.w3.org/WAI/standards-guidelines/aria/) [[WAI-ARIA]], [Scaleable Vector Graphics](https://www.w3.org/TR/SVG/) [[SVG2]], [EPUB 3.1](http://www.idpf.org/epub/31/spec/epub-spec.html), and more. The ACT Rules Format is designed to be technology agnostic, meaning that it can conceivably be used to describe test rules for other technologies.
 
 The ACT Rules Format can be used to describe ACT Rules dedicated to testing the [=accessibility requirements=] defined in Web Content Accessibility Guidelines [[WCAG]], which are specifically designed for web content. Other accessibility requirements applicable to web technologies can also be testable with ACT Rules. For example, ACT Rules could be developed to test the conformance of web-based user agents to the [User Agent Accessibility Guidelines](https://www.w3.org/WAI/standards-guidelines/uaag/) [[UAAG20]]. The ACT Rules Format might not always be suitable to describe tests for other types of accessibility requirements.
 
@@ -35,23 +35,23 @@ ACT Rule Types {#rule-types}
 
 In accessibility, there are often different technical solutions to make the same type of content accessible. For example, there are multiple methods for giving an `img` element in HTML an accessible name. Multiple solutions could be tested in a single rule; however, such a rule tends to be quite complex, making it difficult to understand and maintain. The ACT Rules Format solves this by providing two types of rules:
 
-- <dfn>Atomic rules</dfn> describe how to test a specific type of solution. It contains a precise definition of what elements, nodes or other "parts" of a [=test subject=] are to be tested, and when those elements are considered to fail the rule. These rules are to be kept small and *atomic*. Meaning that atomic rules test a single "failure condition", and do so without using the [=findings=] from other rules.
+- <dfn>Atomic rules</dfn> describe how to test a specific type of solution. It contains a precise definition of what elements, nodes or other "parts" of a [=test subject=] are to be tested, and when the test subject is considered to fail the rule. These rules are to be kept small and *atomic*, this means that atomic rules test a single "failure condition", and do so without using the [=findings=] from other rules.
 
-- <dfn>Composite rules</dfn> describe how the [=findings=] of multiple [=atomic rules=] or composite rules can be combined into a single [=outcome=]. A composite rule can have multiple "satisfying conditions", each of these tested in separate rules. The logic in the composite rule describes how any one of these satisfying conditions, or some combination of them is used to determine a single outcome.
+- <dfn>Composite rules</dfn> describe how the [=findings=] of multiple [=atomic rules=] or composite rules can be combined into a single [=outcome=]. A composite rule can have multiple "satisfying conditions", each of these tested in separate rules. The logic in the composite rule describes how any one of these satisfying conditions, or some combination of them, is used to determine a single outcome.
 
 <div class="example">
   <p>Composite rule: video elements have an audio description or media alternative (<a href="https://www.w3.org/WAI/WCAG21/quickref/#audio-description-or-media-alternative-prerecorded" target="_blank">WCAG 2.1, success criterion 1.2.3 Audio Description or Media Alternative</a>).</p>
-  <p>Each HTML `video` element meets expectations from at least one of the following rules:</p>
+  <p>Each HTML `video` element meets the expectations from at least one of the following rules:</p>
   <ul>
-    <li>video elements have a transcript</li>
-    <li>video elements have an audio description</li>
-    <li>video elements have a description track</li>
+    <li>Video elements have a transcript</li>
+    <li>Video elements have an audio description</li>
+    <li>Video elements have a description track</li>
   </ul>
 </div>
 
-Not all atomic rules have to be part of a composite rule. Composite rules are used when the findings of multiple rules need to be combined in order to determine (non-)conformance of a test subject to an [=accessibility requirement=].
+Not all atomic rules have to be part of a composite rule. Composite rules are used when the [=findings=] of multiple rules need to be combined in order to determine if a [=test subject=] does not satisfy an [=accessibility requirement=].
 
-The separation between atomic rules and composite rules is a division of responsibility. Atomic rules test if web content is correctly implemented in a particular solution. Composite rules can test if a combination of [=findings=] from other rules satisfy the accessibility requirement, in part or as a whole.
+The separation between atomic rules and composite rules creates a division of responsibilities. Atomic rules test if web content is correctly implemented in a particular solution. Composite rules can test if a combination of [=findings=] from other rules satisfy the accessibility requirement, in part or as a whole.
 
 
 ACT Rule Structure {#act-rule-structure}
@@ -64,7 +64,9 @@ An ACT Rule MUST consist of at least the following items:
 * [Rule Description](#rule-description)
 * [Rule type](#rule-types)
 * [Accessibility Requirements Mapping](#accessibility-requirements-mapping)
-* [Input Aspects List](#input-aspects-list) (for atomic rules) OR [Input Rules List](#input-rules-list) (for composite rules)
+* [Rule Input](#input), which is one of the following:
+    * [Input Aspects](#input-aspects) (for atomic rules) OR 
+    * [Input Rules](#input-rules) (for composite rules)
 * [Applicability](#applicability)
 * [Expectations](#expectations)
 * [Assumptions](#assumptions)
@@ -76,16 +78,16 @@ An ACT Rule MUST consist of at least the following items:
 * [Background](#background) (optional)
 * ACT Rules Format [=outcome=] definition
 
-ACT Rules MUST be written in a format that conforms to the Web Content Accessibility Guidelines [[WCAG]] or a comparable accessibility standard. ACT Rule [test cases](#test-cases) are allowed to contain inaccessible content. If any test case contains accessibility issues listed in [WCAG 2.1 Section 5.2.5 Non-Interference](https://www.w3.org/TR/WCAG21/#cc5), users MUST be warned of this in advance. Using an accessible format supports people with disabilities. It also makes internationalization of ACT Rules easier.
+The ACT Rules format does not prescribe what format ACT Rules can be written in (e.g. HTML, DOCX, PDF, etc.). However, ACT Rules MUST be written in a document that conforms to the Web Content Accessibility Guidelines [[WCAG]] or a comparable accessibility standard. This ensures that ACT Rules are accessible to people with disabilities. ACT Rule [test cases](#test-cases) are allowed to contain inaccessible content. If any test case contains accessibility issues listed in [WCAG 2.1 Section 5.2.5 Non-Interference](https://www.w3.org/TR/WCAG21/#cc5), users MUST be warned of this in advance. In addition to supporting people with disabilities, using an accessible format also makes internationalization of ACT Rules easier.
 
 
 Rule Identifier {#rule-identifier}
-==================================
+----------------------------------
 
 An ACT Rule MUST have a unique identifier that can be any unique text, such as plain text, URL or a database identifier.
 
 <div class=example>
-  <p><strong>Example</strong>: ACT Rules may use file names as identifiers. They include a technology directory, followed by a handle that includes an element name or attribute:</p>
+  <p>**Example:** ACT Rules may use file names as identifiers. They include a technology directory, followed by a handle that includes an element name or attribute:</p>
   <ul>
     <li>html+svg/video-alternative</li>
     <li>html+svg/meta-no-refresh</li>
@@ -95,19 +97,19 @@ An ACT Rule MUST have a unique identifier that can be any unique text, such as p
 
 
 Rule Description {#rule-description}
-=========================================
+------------------------------------
 
 An ACT Rule MUST have a description that is in plain language, and provides a brief explanation of what the rule does.
 
 <div class=example>
-  **Example**: This rule checks that the HTML page has a title
+  **Example:** This rule checks that the HTML page has a title
 </div>
 
 
 Accessibility Requirements Mapping {#accessibility-requirements-mapping}
-========================================================
+------------------------------------------------------------------------
 
-When an ACT Rule is designed to test one or more [=Accessibility requirements documents=], the rule MUST list all [=accessibility requirements=] from those documents that are not satisfied when the [=outcome=] of the rule is `failed`. For example, when designing a rule for WCAG 2.1 that tests if image buttons have an alternative text, the rule maps to success criteria 1.1.1 Non-text content, and 4.1.2 Name, Role, Value. That ACT rule will list both success criteria in the accessibility requirements mapping.
+When an ACT Rule is designed to test one or more [=Accessibility requirements documents=], the rule MUST list all [=accessibility requirements=] from those documents that are not satisfied when the [=outcome=] of the rule is `failed`. For example, when designing a rule for WCAG 2.1 that tests if image buttons have an alternative text, the rule maps to success criteria 1.1.1 Non-text content, and 4.1.2 Name, Role, Value. That ACT Rule will list both success criteria in its accessibility requirements mapping.
 
 Each [=accessibility requirement=] in the mapping MUST include the following:
 
@@ -116,19 +118,22 @@ Each [=accessibility requirement=] in the mapping MUST include the following:
 3. a link to the [=accessibility requirements document=] if one exists, and
 4. the conformance level associated with the accessibility requirement, if one exists.
 
+
+### Outcome Mapping ### {#outcome-mapping}
+
 For each accessibility requirement in the mapping, an ACT Rule MUST indicate what the [=outcome=] of the rule means for satisfying that accessibility requirement. When the outcome is `failed`, the accessibility requirements are *<dfn>not satisfied</dfn>*. When the outcome is `passed` or `inapplicable`, the accessibility requirements could be *<dfn>satisfied</dfn>*, or *<dfn>further testing could be necessary</dfn>*. Rules that can be used to determine if an accessibility requirement is *satisfied* are called <dfn>satisfying tests</dfn>.
 
 <div class=note>
-**Note**: In WCAG 2.1, success criteria do not evaluate to `passed`, `failed` or `inapplicable`. Rather they can be *satisfied* (or not). (See the [WCAG 2.1 definition: satisfies a success criterion](https://www.w3.org/TR/WCAG21/#dfn-satisfies).) If a success criterion is *not satisfied* a web page can only conform if there is a conforming alternative version, as described in [WCAG 2.1 Conformance Requirement 1](https://www.w3.org/TR/WCAG21/#cc1).
+**Note:** In WCAG 2.1, success criteria do not evaluate to `passed`, `failed` or `inapplicable`. Rather they can be *satisfied* (or not). (See the [WCAG 2.1 definition: satisfies a success criterion](https://www.w3.org/TR/WCAG21/#dfn-satisfies).) If a success criterion is *not satisfied* a web page can only conform if there is a conforming alternative version, as described in [WCAG 2.1 Conformance Requirement 1](https://www.w3.org/TR/WCAG21/#cc1).
 </div>
 
 <div class=example>
-  <p><strong>Example:</strong> Accessibility Requirements Mapping for a rule that tests if an image button has an accessible name:
+  <p>**Example:** Accessibility Requirements Mapping for a rule that tests if an image button has an accessible name:
   <blockquote><ul>
-    <li><a href="https://www.w3.org/TR/WCAG21/#non-text-content">
-        Success Criterion 1.1.1: Non-text content
-      </a><ul>
-        <li><strong>Required for conformance</strong> to WCAG 2.0 and WCAG 2.1 level A</li>
+    <li>
+      [Success Criterion 1.1.1: Non-text content](https://www.w3.org/TR/WCAG21/#non-text-content)
+      <ul>
+        <li>**Required for conformance** to WCAG 2.0 and WCAG 2.1 level A</li>
         <li>Outcome mapping: <ul>
           <li>Failed outcome: not satisfied</li>
           <li>Passed outcome: further testing is needed</li>
@@ -136,10 +141,10 @@ For each accessibility requirement in the mapping, an ACT Rule MUST indicate wha
         </ul></li>
       </ul>
     </li>
-    <li><a href="https://www.w3.org/TR/WCAG21/#name-role-value">
-        Success Criterion 4.1.2: Name, Role, Value
-      </a><ul>
-        <li><strong>Required for conformance</strong> to WCAG 2.0 and WCAG 2.1 level A</li>
+    <li>
+      [Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value)
+      <ul>
+        <li>**Required for conformance** to WCAG 2.0 and WCAG 2.1 level A</li>
         <li>Outcome mapping:<ul>
           <li>Failed outcome: not satisfied</li>
           <li>Passed outcome: further testing is needed</li>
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -17,43 +17,41 @@ Introduction {#intro}
...
+- <dfn>Composite rules</dfn> describe how the [=findings=] of multiple [=atomic rules=] or composite rules can be combined into a single [=outcome=]. A composite rule can have multiple "satisfying conditions", each of these tested in separate rules. The logic in the composite rule describes how any one of these satisfying conditions, or some combination of them is used to determine a single outcome.
 
 <div class="example">
-  <p>Composite rule: Header cells in HTML tables (<a href="https://www.w3.org/WAI/WCAG21/quickref/#info-and-relationships" target="_blank">WCAG 2 Success Criterion 1.3.1</a>).</p>
-  <p>This rule uses outcomes from the following rules:</p>
+  <p>Composite rule: video elements have an audio description or media alternative (<a href="https://www.w3.org/WAI/WCAG21/quickref/#audio-description-or-media-alternative-prerecorded" target="_blank">WCAG 2.1, success criterion 1.2.3 Audio Description or Media Alternative</a>).</p>
+  <p>Each HTML `video` element meets expectations from at least one of the following rules:</p>
```
nitedog: Suggested edit:

> Each HTML `video` element meets the expectations defined by at least one of the following rules:

Rationale: "meets expectations" is vague - it could mean meeting only parts of the defined expectation.

act-rules-format.bs
```diff
@@ -156,7 +150,7 @@ For each accessibility requirement in the mapping, an ACT Rule MUST indicate wha
...
+ACT Rules can be used to test accessibility requirements that are not part of a W3C accessibility standard, such as [Hypertext Markup Language](https://www.w3.org/TR/html/) [[HTML]] requirements, or a testing methodology like [RGAA 3 2016](https://disic.github.io/rgaa_referentiel_en/criteria.html). An ACT Rule MUST indictate whether or not the [=accessibility requirement=] it maps to is required for conformance in its [=accessibility requirements document=]. Examples of accessibility requirements that are not required for conformance are WCAG sufficient techniques, or a company style guide which includes both requirements and optional "best practices". The distinction between what is required and what is optional has to be made clear.
```
nitedog: "The distinction ... has to be made clear" - should that be a MUST requirement?

act-rules-format.bs
```diff
@@ -188,7 +182,7 @@ ACT Rules can be used to test accessibility requirements that are not part of a
 
 If the `failed` outcome can not be mapped to *not satisfied* for an accessibility requirement, that requirement MUST NOT be included in the accessibility requirements mapping. The optional [Background section](#background) could be used to list accessibility requirements and standards that may be thematically related, but for which the rule is not a failure test.
```
nitedog: Misuse of "may" keyword. Here some suggestions:

- could be used to list accessibility requirements and standards that are thematically related (remove "may")
- could be used to list accessibility requirements and standards when they are thematically related (replace "may")
- MAY be used to list accessibility requirements and standards that are thematically related (move "may" to make it a requirement)

act-rules-format.bs
```diff
@@ -198,20 +192,27 @@ If the rule does not map to any [=accessibility requirement=], the accessibility
...
+**Note**: While rules are often designed for one, or possibly a small collection of [=accessibility requirements document=], it is likely that other accessibility requirements documents also map to those rules. For example, rules may be written for WCAG 2.1, but many of those may map to a company's internal accessibility policy. In such a scenario, an <dfn>external accessibility requirements mapping</dfn> could be created. An external accessibility requirements mapping ammend the accessibility requirements mapping of a rule by adding mapping to a different accessibility requirements document. An external accessibility requirements mapping avoids duplication of a rule, for the sole purpose of changing the mapping.
```
nitedog: Replace "may" with "can" in the example.

nitedog: Consider making "could be created" to "MAY be created" (ie. changing it to a requirement).

act-rules-format.bs
```diff
@@ -17,43 +17,41 @@ Introduction {#intro}
...
+- <dfn>Composite rules</dfn> describe how the [=findings=] of multiple [=atomic rules=] or composite rules can be combined into a single [=outcome=]. A composite rule can have multiple "satisfying conditions", each of these tested in separate rules. The logic in the composite rule describes how any one of these satisfying conditions, or some combination of them is used to determine a single outcome.
 
 <div class="example">
-  <p>Composite rule: Header cells in HTML tables (<a href="https://www.w3.org/WAI/WCAG21/quickref/#info-and-relationships" target="_blank">WCAG 2 Success Criterion 1.3.1</a>).</p>
-  <p>This rule uses outcomes from the following rules:</p>
+  <p>Composite rule: video elements have an audio description or media alternative (<a href="https://www.w3.org/WAI/WCAG21/quickref/#audio-description-or-media-alternative-prerecorded" target="_blank">WCAG 2.1, success criterion 1.2.3 Audio Description or Media Alternative</a>).</p>
```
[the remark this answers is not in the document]
moekraft: Should video be uppercase? Composite rule: Video elements have an audio...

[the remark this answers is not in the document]
annethyme: Decided in the ACT TF meeting that this is not relevant for the examples.

act-rules-format.bs
WilcoFiers: "meets all expectations"

act-rules-format.bs
WilcoFiers: editorial

act-rules-format.bs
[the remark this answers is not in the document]
WilcoFiers: Editorial, will look at it.

act-rules-format.bs
```diff
@@ -198,20 +192,27 @@ If the rule does not map to any [=accessibility requirement=], the accessibility
...
+**Note**: While rules are often designed for one, or possibly a small collection of [=accessibility requirements document=], it is likely that other accessibility requirements documents also map to those rules. For example, rules may be written for WCAG 2.1, but many of those may map to a company's internal accessibility policy. In such a scenario, an <dfn>external accessibility requirements mapping</dfn> could be created. An external accessibility requirements mapping ammend the accessibility requirements mapping of a rule by adding mapping to a different accessibility requirements document. An external accessibility requirements mapping avoids duplication of a rule, for the sole purpose of changing the mapping.
 </div>
 
 
-Aspects Under Test (Atomic rules only) {#input-aspects}
-=======================================================
+Rule Input {#input}
```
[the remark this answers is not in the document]
WilcoFiers: We can use a sublist for it in the rule structure

act-rules-format.bs
WilcoFiers: Going with "meets the expectations" as suggested. "All expectations from one rule" is a bit confusing IMO. I'm not using "as defined in", as I don't think it helps readability. This is an example, making this editorial. We can revisit it if we have to in CR.

act-rules-format.bs
```diff
@@ -105,15 +103,11 @@ An ACT Rule MUST have a description that is in plain language, and provides a br
   **Example**: This rule checks that the HTML page has a title
```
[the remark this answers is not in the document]
WilcoFiers: I disagree. If the test subject is a single page, you wouldn't talk about "each HTML document". Similarly, we don't say "Each img element in each HTML document". You'd use that language if you're talking about a set of web pages. So for example: "The rule checks that the title of an HTML page is unique within a set of web pages".

act-rules-format.bs
```diff
@@ -296,50 +296,48 @@ Note that atomic rules in a composite rule MAY have different applicability. Bec
...
+An atomic rule expectation MUST only use information available in the [input aspects](#input-aspects-list), from the applicability, and other expectations of the same rule. No other information can be used in the expectation. So for instance, an expectation could be "Expectation 1 is true and ...", but it can't be "Rule XYZ passed and ...". This ensures that atomic rules are encapsulated.
 
 <div class=example>
   <p><strong>Example</strong>: A rule for labels of HTML `input` elements might have the following expectations:</p>
   <ol>
-    <li>The test target has an accessible name (as described in [Accessible Name and Description: Computation and API Mappings 1.1](https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_te)). [[accname-aam-1.1]]</li>
-    <li>The accessible name describes the purpose of the test target.</li>
+    <li>Each test target has an accessible name (as described in [Accessible Name and Description: Computation and API Mappings 1.1](https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_te)). [[accname-aam-1.1]]</li>
```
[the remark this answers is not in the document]
WilcoFiers: All rules with multiple expectations can be split. What makes you suggest specifically this one needs to change? What rule would you suggest?

act-rules-format.bs
```diff
@@ -296,50 +296,48 @@ Note that atomic rules in a composite rule MAY have different applicability. Bec
...
 
-A [=composite rule=] MUST contain one or more expectations that describes the logic that is used to determine a single `passed` or `failed` [=outcome=] for each [=test target=], based on the outcomes of [=atomic rules=] or [=composite rules=] listed in its [input rules list](#input-rules-list).
+When atomic rules are used in composite rules, the outcome of the composite rules can be `passed` even if the outcome of one of its input rules is `failed`, depending on the [expectations](#expectations-composite) of the composite rule.
 
-When all expectations are true for a test target, the test target `passed` the rule. If one or more expectations is false, the test target `failed` the rule. This works the same way for atomic rules. A composite rule expectation MUST NOT use information from [test aspects](#input-aspects).
+All expectations of a [=composite rule=] MUST describe the logic that is used to determine a single `passed` or `failed` [=outcome=] for a [=test subject=], based on the [=findings=] of rules in its [input rules list](#input-rules-list). A composite rule expectation MUST NOT use information from [input aspects](#input-aspects-list).
```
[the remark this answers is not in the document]
WilcoFiers: Correct. We've defined outcome in relationship to test subject, not to test targets.

act-rules-format.bs
```diff
@@ -296,50 +296,48 @@ Note that atomic rules in a composite rule MAY have different applicability. Bec
...
-  <p>A composite rule for <a href="https://www.w3.org/WAI/WCAG21/quickref/#audio-description-or-media-alternative-prerecorded" target="_blank">WCAG 2.1 Audio Description or Media Alternative</a> applies a logic across results from three atomic rules. The expectation of the composite rule is as follows:</p>
-  <blockquote>
-    <p>For each test target, the outcome of one of the following rules is `passed`:</p>
-    <ul>
-      <li>Video elements have an audio description (video-audio-desc)</li>
-      <li>Video elements have a media alternative (video-text-alt)</li>
-    </ul>
-  </blockquote>
+  <p>Composite rule: video elements have an audio description or media alternative (<a href="https://www.w3.org/WAI/WCAG21/quickref/#audio-description-or-media-alternative-prerecorded" target="_blank">WCAG 2.1, success criterion 1.2.3 Audio Description or Media Alternative</a>).</p>
+  <p>Each HTML `video` element meets expectations from at least one of the following rules:</p>
```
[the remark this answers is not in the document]
WilcoFiers: As above.

moekraft:
Under Rule Structure, I think we should have a section for Outcome rather than a link to ACT Rules Format outcome definition which is static. This differs from Romain's suggestion in Issue #284

annethyme:
As discussed in our last ACT TF teleconference meeting, Siteimprove was not aware that we were looking at changing the scope of outcome reporting for individual rules from reporting at the **test target** level to reporting at the **test subject** level. This change means that the rules should be reporting a “passed”/”failed” outcome for a whole page at a time instead of reporting back which specific elements/attributes/things are passing or failing the rule.
 
The result of this change, as we see it, are less granular results that will not give the same level of comparability across tools and methodologies that we hoped to get from the ACT Rules Format.
 
Our suggestion is to have an outcome mapping where either the result is “inapplicable” – no test targets were found within the test subject, or a “passed”/”failed” outcome for each test target. We don’t see a problem with this “mix”, that seems quite logical to us.
 
We thought that this was what we were getting, since closely reading the rules format said “test target” everywhere, and no issues or agenda items have mentioned that we were looking at switching “test target” out with “test subject” level reporting. This is the only reason we have not argued for this previously.
 
It seemed on the call last week that we might be the only ones with this concern though, so if we cannot revert to the previous wording, we would like to suggest that as a compromise, we mention in the ACT Rules Format, that reporting a passed/failed outcome per test subject is recommended to ensure transparency and comparability of results.

act-rules-format.bs
```diff
@@ -245,183 +259,212 @@ For other aspects that are not well defined, an ACT Rule MUST include either a d
...
+A [=composite rule=] defines how the [=findings=] from rules in its [input rules](#input-rules) are used to determine a single [=outcome=].
 
-The applicability of a composite rule is defined as the union of all the applicability sections of its [=atomic rules=] as well as atomic rules that make up any composite rules in the [input rules list](#input-rules-list). Rule authors MAY describe the applicability for composite rules. This can be useful if it is difficult to express the combined applicability in plain language. If the composite rule includes applicability, it MUST be the union of all the applicability sections of the atomic rules.
+The applicability of a composite rule is defined as the union of all the applicability sections of rules in the [input rules](#input-rules). This includes [=atomic rules=], and [=composite rules=] which themselves have a union applicability. Rule authors MAY ommit a description of the applicability for composite rules. This can be useful if it is difficult to express the combined applicability in plain language. If the composite rule includes applicability, it MUST be the union of all the applicability in the [input rules](#input-rules).
```
annethyme: Editorials needed for "it MUST be the union of all the applicability in the [input rules]". Maybe add "sections" to "applicability"

act-rules-format.bs
```diff
@@ -245,183 +259,212 @@ For other aspects that are not well defined, an ACT Rule MUST include either a d
...
 
-Note that atomic rules in a composite rule MAY have different applicability. Because of this, not every element applicable within the composite rule is tested by every atomic rule.
+Note that input rules in a composite rule MAY have different applicability. Because of this, not every element applicable within the composite rule is tested by every input rule.
```
annethyme: element --> test target?

act-rules-format.bs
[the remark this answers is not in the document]
annethyme: I still say that it has to meet all expectations of at least one input rule. It can't meet half an input rule and still pass.

annethyme:
If the change of outcome reporting from the test target to the test subject level came out of the discussion at TPAC on how to accommodate possible future accessibility requirement that has the form of e.g. "85% of images have a text alternative", I would say that this could also be done in a format where the outcome is related to the test target. 

Example: 

> Applicability: This rule applies to the full set of `img` elements in a HTML document.
> Expectation: 85% of the elements in the applicable set have an accessible name. 

Example of an Auto-WCAG draft rule that is treating a set of elements: https://github.com/auto-wcag/auto-wcag/pull/220/files.

act-rules-format.bs
```diff
@@ -451,63 +494,74 @@ Definitions {#definitions}
...
     </div>
     <div class=note>
-      <p><strong>Note</strong>: Implementers using the [[EARL10-Schema]] can express the test subject with the [subject property](https://www.w3.org/TR/EARL10-Schema/#subject)</p>
+      <p>**Note:**>: Implementers using the [[EARL10-Schema]] can express the test subject with the [subject property](https://www.w3.org/TR/EARL10-Schema/#subject)</p>
     </div>
   </dd>
 
   <dt><dfn>Test Target</dfn></dt>
   <dd>
     <p>A distinct part of the [=test subject=].</p>
```
annethyme: Should it say here that the test targets are the things that are applicable to the rule?
I know that it becomes a bit circular when following the link from the Applicability, but out of context, the definition seems very thin.

act-rules-format.bs
```diff
@@ -245,183 +259,212 @@ For other aspects that are not well defined, an ACT Rule MUST include either a d
...
 
 <div class=example>
-  <p><strong>Example</strong>: The applicability of an atomic rule testing <a href="https://www.w3.org/WAI/WCAG21/quickref/#audio-control" target="_blank">WCAG 2.1 Audio Control:</a></p>
+  <p>**Example:** The applicability of an atomic rule testing [WCAG 2.1 Audio Control](https://www.w3.org/WAI/WCAG21/quickref/#audio-control):</a></p>
   <blockquote>
-    <p>Any <code>video</code> or <code>audio</code> element(s) with the <code>autoplay</code> attribute, as well as any <code>object</code> element(s) that is used for automatically playing video or audio when the web page loads.</p>
+    <p>Each `video` or `audio` element with the `autoplay` attribute, as well as each `object` element that is used for automatically playing video or audio when the web page loads.</p>
+    <p>**Note: A web page is considerd "loaded" when the `document.readySate` is set to `complete`.</p>
```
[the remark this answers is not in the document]
WilcoFiers: Not if we allow inline definitions like you suggested before. That's a good addition.

act-rules-format.bs
```diff
@@ -15,83 +15,83 @@ Markup Shorthands: markdown yes
...
+- <dfn>Composite rules</dfn> describe how the [=findings=] of multiple [=atomic rules=] or composite rules can be combined into a single [=outcome=]. A composite rule can have multiple "satisfying conditions", each of these tested in separate rules. The logic in the composite rule describes how any one of these satisfying conditions, or some combination of them, is used to determine a single outcome.
 
 <div class="example">
-  <p>Composite rule: Header cells in HTML tables (<a href="https://www.w3.org/WAI/WCAG21/quickref/#info-and-relationships" target="_blank">WCAG 2 Success Criterion 1.3.1</a>).</p>
-  <p>This rule uses outcomes from the following rules:</p>
+  <p>Composite rule: video elements have an audio description or media alternative (<a href="https://www.w3.org/WAI/WCAG21/quickref/#audio-description-or-media-alternative-prerecorded" target="_blank">WCAG 2.1, success criterion 1.2.3 Audio Description or Media Alternative</a>).</p>
+  <p>Each HTML `video` element meets the expectations from at least one of the following rules:</p>
```
nitedog: suggest changing "meets the expectations" to "meets all expectations" for further emphasis, and I think also smoother reading.

nitedog: I suggest changing "meets the expectations" to "meets all expectations", to emphasize the meaning and for better readability ("the" implies a definite set, which makes me pause to think which set you actually mean. On the other hand, "all" implies the entire set and does not require further processing cylces in my view).

act-rules-format.bs
```diff
@@ -101,23 +101,19 @@ An ACT Rule MUST have a unique identifier that can be any unique text, such as p
...
-========================================================
+------------------------------------------------------------------------
 
-When an ACT Rule is designed to test the conformance to one or more [=Accessibility requirements documents=], the rule MUST list all [=accessibility requirements=] from those documents that are not satisfied when the outcome of the rule is `failed`. For example, when designing a rule for WCAG 2.1 that tests if image buttons have an alternative text, the rule maps to success criteria 1.1.1 Non-text content, and 4.1.2 Name, Role, Value.
+When an ACT Rule is designed to test one or more [=Accessibility requirements documents=], the rule MUST list all [=accessibility requirements=] from those documents that are not satisfied when the [=outcome=] of the rule is `failed`. For example, when designing a rule for WCAG 2.1 that tests if image buttons have an alternative text, the rule maps to success criteria 1.1.1 Non-text content, and 4.1.2 Name, Role, Value. That ACT Rule will list both success criteria in its accessibility requirements mapping.
 
 Each [=accessibility requirement=] in the mapping MUST include the following:
```
nitedog: In the item below, add "identifier" to "name, title, or summary"

act-rules-format.bs
```diff
@@ -126,30 +122,33 @@ Each [=accessibility requirement=] in the mapping MUST include the following:
 3. a link to the [=accessibility requirements document=] if one exists, and
```
nitedog: change "link" to "a link or reference" (eg. if the document is not publicly available or directly linkable for any other reason)

act-rules-format.bs
nitedog: I think the issue is with the "the" in the sentence. How about "all applicability definitions" instead of "sections"? The grammar of the sentence does seem wrong to me too.

act-rules-format.bs
```diff
@@ -245,183 +259,212 @@ For other aspects that are not well defined, an ACT Rule MUST include either a d
...
 
-The expectations describe what the requirements are for the [=test targets=] derived from the [applicability](#applicability).
+Each expectation MUST be distinct, unambiguous, and be written in plain language. Unlike in applicability, a certain level of subjectivity is allowed in expectations. Meaning that the expectation has only one possible meaning, but that meaning MAY not fully quantifiable.
 
+Commonly, an expectation comes in the form of an "Each test target has ..." statement. In such an expectation, all test targets need to [=meet expectations|meet all expectations=] for the rule to pass. It is also possible to have an expectation where not all test targets meet all expectations. For example: "85% of all test targets have an accessible name" is an expectation that is true even when 15% of the test targets do not meet the expectations.
```
nitedog: (important) I think this paragraph needs to be rewritten, to better explain the different types of definition forms for expectations. It seems that we have two types:

1. individual test targets - the expectation applies to individual test targets, like "every test target has an accessible name"
2. sets of test targets - the expectation applies to a set of test targets, like "85% of all test targets have an accessible name"

act-rules-format.bs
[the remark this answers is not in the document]
nitedog: Since this is only an example, could we raise an issue for it and come back to this after CR publication?

act-rules-format.bs
```diff
@@ -512,13 +512,33 @@ Definitions {#definitions}
...
+    <div class=example>
+      <p>**Example:** An HTML `img` element an `alt` attribute meet the following expectation:</p>
+      <blockquote>Each test target has an alt attribute</blockquote>
+    </div>
+    <div class=example>
+      <p>**Example:** An HTML element with a duplicate ID does not meet the following expectation:</p>
+      <blockquote>No test target has the same ID</blockquote>
+    </div>
+    <div class=example>
+      <p>**Example:** An HTML `video` element without captions does not meet the following expectation:</p>
```
annethyme: How do you know if this video is in the 80% or 20%?

act-rules-format.bs
```diff
@@ -512,13 +512,33 @@ Definitions {#definitions}
...
+  </dd>
+
   <dt><dfn>Outcome</dfn></dt>
   <dd>
     <p>One of three types of conclusions that come from evaluating an ACT Rule on a [=test subject=]. An outcome can be:<p>
     <ul>
       <li>**Inapplicable:** No part of the test subject matches the applicability</li>
-      <li>**Passed:** The test subject meets all expectations</li>
-      <li>**Failed:** The test subject does not meet all expectations</li>
+      <li>**Passed:** All expectations are true for the test subject</li>
```
annethyme: Shouldn't we se the "meets all expectations" definition here?

act-rules-format.bs
annethyme: ... building on what Shadi said, I think that the "85%..." type of expectation rules out saying anything about whether a particular test target meets the expectation. These types of rules can only say something about the set they apply to. 

(Even though an implementation could choose to report which elements count toward/from meeting the expectation. And it would probably be nice to report some calculations too, e.g. how many % had an accessible name, and how far is it from the 85% threshold).

I also don't really see why it's not always "meet **all** expectations". I agree that it's not always "all test targets", like in the 85% case, but I cannot imagine an example where we would allow for only meeting e.g. 2 out of 3 expectations in a rule. If the expectation is "85%...", this is still an expectation that needs to be met.

act-rules-format.bs
```diff
@@ -15,83 +15,83 @@ Markup Shorthands: markdown yes
...
 
 
 ACT Rule Types {#rule-types}
 ============================
 
-In accessibility, there are often different technical solutions to make the same type of content accessible. Multiple solutions could be tested in a single rule; however, such a rule tends to be quite complex, making it difficult to understand and maintain. The ACT Rules Format solves this by providing two types of rules:
+In accessibility, there are often different technical solutions to make the same type of content accessible. For example, there are multiple methods for giving an `img` element in HTML an accessible name. Multiple solutions could be tested in a single rule; however, such a rule tends to be quite complex, making it difficult to understand and maintain. The ACT Rules Format solves this by providing two types of rules:
```
maryjom: Editorial: Suggest added sentence read as follows...
For example, there are multiple methods to give an accessible name to an HTML 'img' element.

act-rules-format.bs
```diff
@@ -15,83 +15,83 @@ Markup Shorthands: markdown yes
...
 
-- <dfn>Atomic rules</dfn> describe how to test a specific type of solution. It contains a precise definition of what elements, nodes or other parts of a [=test subject=] are to be tested, and when those elements are considered to fail the rule. These rules are to be kept small and *atomic*. Meaning that atomic rules test a single "failure condition", and do so without using the [=outcome=] of other rules.
+- <dfn>Atomic rules</dfn> describe how to test a specific type of solution. It contains a precise definition of what elements, nodes or other "parts" of a [=test subject=] are to be tested, and when the test subject is considered to fail the rule. These rules are to be kept small and *atomic*. This means that atomic rules test a single "failure condition", and do so without using the [=findings=] from other rules.
```
maryjom: I'm not sure I understand the reason for changing "outcome" to "findings". The "outcome" is the finding of an ACT rule, whether it be atomic or composite. I don't think we need two different terms for that.

act-rules-format.bs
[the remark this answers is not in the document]
maryjom: > This is both a WCAG 2.0 and 2.1 requirement. How do we expect people to communicate this? And should we show this in the examples?

I don't think rules should have to quote every single version that a requirement appears in or you'll wind up with a very long list.  Maybe simply have the standard without the version and a link pointer would be best.

act-rules-format.bs
```diff
@@ -15,83 +15,83 @@ Markup Shorthands: markdown yes
...
-    <li>Header indicated by using ARIA labels</li>
+    <li>Video elements have a transcript</li>
+    <li>Video elements have an audio description</li>
+    <li>Video elements have a description track</li>
   </ul>
-  <p>If any one of these rules passes, the outcome of the composite rule for the table cell is "passed".</p>
 </div>
 
-The separation between atomic rules and composite rules can be seen as a division of responsibility. Atomic rules typically test if web content is correctly implemented in a particular solution. Composite rules can test if a combination of outcomes from other rules satisfy the accessibility requirement, in part or as a whole.
+Not all atomic rules have to be part of a composite rule. Composite rules are used when the [=findings=] of multiple rules need to be combined in order to determine if a [=test subject=] does not satisfy an [=accessibility requirement=].
```
maryjom: Editorial: Remove "in order"

act-rules-format.bs
```diff
@@ -15,83 +15,83 @@ Markup Shorthands: markdown yes
...
+The ACT Rules format does not prescribe what format ACT Rules can be written in (e.g. HTML, DOCX, PDF, etc.). However, ACT Rules MUST be written in a document that conforms to the Web Content Accessibility Guidelines [[WCAG]] or a comparable accessibility standard. This ensures that ACT Rules are accessible to people with disabilities. ACT Rule [test cases](#test-cases) are allowed to contain inaccessible content. If any test case contains accessibility issues listed in [WCAG 2.1 Section 5.2.5 Non-Interference](https://www.w3.org/TR/WCAG21/#cc5), users MUST be warned of this in advance. In addition to supporting people with disabilities, using an accessible format also makes internationalization of ACT Rules easier.
```
maryjom: Where should accessibility issues in test cases be documented to meet this "warned in advance" requirement?  In the test case itself, in the rule description? Is that flexible?

maryjom: Also, there is the use of normative language - "MUST" and this is a note, right?  Is this requirement in the ACT Rules Format somewhere?

act-rules-format.bs
nitedog: See issue #327 on removing "failure", and simply say "condition" instead.

act-rules-format.bs
```diff
@@ -15,83 +15,83 @@ Markup Shorthands: markdown yes
...
 
-- <dfn>Composite rules</dfn> describe how the [=outcomes=] of multiple [=atomic rules=] or composite rules can be combined into a single outcome. A composite rule can have multiple "satisfying conditions", each of these provided in separate atomic or composite rules. The logic in the composite rule describes how any one of these satisfying conditions, or some combination of them is used to determine a single outcome.
+- <dfn>Composite rules</dfn> describe how the [=findings=] of multiple [=atomic rules=] or composite rules can be combined into a single [=outcome=]. A composite rule can have multiple "satisfying conditions", each of these tested in separate rules. The logic in the composite rule describes how any one of these satisfying conditions, or some combination of them, is used to determine a single outcome.
```
nitedog: Maybe remove "satisfying" from here (just "conditions", as suggested in issue #327)?

act-rules-format.bs
nitedog: Following the ACT TF discussion on 31 January 2019, I suggest we also change "Each HTML 'video' element" to "All HTML 'video' elements", to better emphasize the scope.

act-rules-format.bs
nitedog: To your latter comment @maryjom, AFAIK this is not a note but a paragraph in a normative section of the document. It actually sets two normative requirements (accessible content and advance warning).

act-rules-format.bs
```diff
@@ -245,183 +259,212 @@ For other aspects that are not well defined, an ACT Rule MUST include either a d
...
 
 <div class=example>
-  <p><strong>Example</strong>: The applicability of an atomic rule testing <a href="https://www.w3.org/WAI/WCAG21/quickref/#audio-control" target="_blank">WCAG 2.1 Audio Control:</a></p>
+  <p>**Example:** The applicability of an atomic rule testing [WCAG 2.1 Audio Control](https://www.w3.org/WAI/WCAG21/quickref/#audio-control):</a></p>
   <blockquote>
-    <p>Any <code>video</code> or <code>audio</code> element(s) with the <code>autoplay</code> attribute, as well as any <code>object</code> element(s) that is used for automatically playing video or audio when the web page loads.</p>
+    <p>Each `video` or `audio` element with the `autoplay` attribute, as well as each `object` element that is used for automatically playing video or audio when the web page loads.</p>
```
nitedog: Following the ACT TF discussion on 31 January 2019, Please change "Each 'video' or 'audio' element ... as well as each 'object' element" to "All 'video' and 'audio' elements ... as well as all 'object' elements"

act-rules-format.bs
```diff
@@ -245,183 +259,212 @@ For other aspects that are not well defined, an ACT Rule MUST include either a d
...
     <p>All `img` elements.</p>
   </blockquote>
 </div>
 
 
 Expectations {#expectations}
-==============================
+----------------------------
+
+An ACT Rule MUST contain one or more expectations. The expectations describe what the requirements are for [=test targets=] derived from the [applicability](#applicability). An expectation is an assertion about all [=test targets=] that is true if the [=test subject=] satisfies the [=accessibility requirement=]. When all expectations are true for a test subject, the test subject `passed` the rule. If one or more expectations are false, the test subject `failed` the rule. If there are no test targets, the [=outcome=] for the test subject is `inapplicable`.
```
nitedog: Following the ACT TF discussion on 31 January 2019, I suggest we change this paragraph as follows:

> An ACT Rule MUST contain one or more expectations. The expectations describe requirements for the set of [=test targets=] derived from the [applicability](#applicability). An expectation is an assertion about the entire set of all [=test targets=], and is true when the [=test subject=] satisfies the [=accessibility requirement=]. When all expectations are true for a test subject, the test subject `passed` the rule. If one or more expectations are false, the test subject `failed` the rule. If there are no test targets, the [=outcome=] for the test subject is `inapplicable`.

nitedog: Are we certain about "that is true if the test subject satisfies the accessibility requirement"? In think that individual expectations in individual atomic rules may be not met in a composite rule, while the test subject overall satisfies the accessibility requirement.

nitedog: Following the ACT TF discussion on 31 January 2019, I think we should rather say:

> An expectation is an assertion about all test targets or about the collective set of test targets.

act-rules-format.bs
```diff
@@ -245,183 +259,212 @@ For other aspects that are not well defined, an ACT Rule MUST include either a d
...
   </ol>
 </div>
 
-An atomic rule expectation MUST only use information available in the [test aspects](#input-aspects), from the applicability, and other expectations of the same rule. No other information can be used in the expectation. So for instance, an expectation could be "Expectation 1 is true and ...", but it can't be "Rule XYZ passed and ...". This ensures the rule is encapsulated.
 
+### Expectations for Composite Rules ### {#expectations-composite}
 
-Expectations for Composite Rules {#expectations-composite}
---------------------------------------------------------
+All expectations of a [=composite rule=] MUST describe the logic that is used to determine a single `passed` or `failed` [=outcome=] for a [=test subject=], based on the [=findings=] of rules in its [input rules](#input-rules). The outcome for a test target is `inapplicable` when all input rules are inapplicable.
```
nitedog: I think the sentence "All expectations of a [=composite rule=] MUST describe" should be changed to "The set of all expectations of a [=composite rules=] MUST collectively describe", because not every expectation must result in a single passed or failed but rather the collective set of one or more expectations.

act-rules-format.bs
```diff
@@ -451,63 +494,94 @@ Definitions {#definitions}
...
+    <ul>
+      <li>All [=test targets=] in the test subject</li>
+      <li>Which test targets met which [expectations](#expectations) in the test subject</li>
+      <li>The [=outcome=] (passed, failed or inapplicble) for the test subject</li>
+    </ul>
+  </dd>
+
+  <dt><dfn>meet expectations</dfn></dt>
+  <dd>
+    <p>A [=test target=] meets an expectation, when that expectation is true for the test target individually, or when the test target increases a count that is required for the expectation to be true.</p>
```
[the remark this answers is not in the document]
nitedog: Following the ACT TF discussion on 31 January 2019 and the suggested changes above, I propose we change this definition to:

> An expectation is met when it is true for the individual test target or set of test targets being tested.

act-rules-format.bs
nitedog: Following the ACT TF discussion on 31 January 2019, I suggest we change this example to:

> A set of 3 HTML 'video' elements of which 2 do not have captions does not meet the following expectation:

act-rules-format.bs
nitedog: I'm wondering why we are distinguishing "meets" and "is true".

act-rules-format.bs
WilcoFiers: Proposal:
> An expectation is an assertion about a [=test target=]. This assertion is true when the [=test subject=] satisfies the [=accessibility requirement=]. When a test target meets all expectations, the test target `passed` the rule. If the test target does not meet one or more expectations, the test target `failed` the rule. If there are no test targets, the [=outcome=] for the rule is `inapplicable`.

act-rules-format.bs
```diff
@@ -451,63 +494,94 @@ Definitions {#definitions}
...
+    <div class=example>
+      <p>**Example:** An HTML `video` element without captions does not meet the following expectation:</p>
+      <blockquote>80% of all test targets has a caption</blockquote>
+    </div>
+    <div class=note>
+      <p>Expectations are true or false assertions about a test subject. Individual test targets can meet all expectations, even if some expectations are false about the test subject as a whole</p>
+    </div>
+  </dd>
+
+  <dt><dfn>Outcome</dfn></dt>
```
WilcoFiers: **Note:** A rule has one passed of failed outcome for every test target. When there are no test targets the rule has one inappliable outcome. This means that each test subject will have one or more outcomes.

WilcoFiers:
PR has been superseded.

PR #346, opened 2019-03-20
edit: Add 2 applicability examples

### LINKED ISSUE #288: Add note that applicability can include the page as a whole or the root node of a document

### LINKED ISSUE #298: Add examples to Applicability
Purpose: Help rule authors make more informed choices for their applicability.

E.g. 
- attribute level
- element level 
- page level
- page collection level

Maybe also consider including something on remember to not only think about which elements to test, but in which scenario, e.g. if they are exposed visually, to screen readers, focusable, etc.

### LINKED ISSUE #330: Add an example or two of applicability
Look at auto-wcag rule to see if we have another good example to add.

WilcoFiers:
Closes #288
Closes #298
Closes #330

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -296,6 +296,19 @@ Even concepts like headings and images can be misunderstood. These terms could r
   </blockquote>
 </div>
 
+<div class=example>
+  <p>**Example:** The applicability of a rule with the page as a test target</a></p>
+  <blockquote>
+    <p>The rule applies to any page where the root element is an `html` element, and which is not embedded in another page.</p>
+  </blockquote>
+</div>
+
+<div class=example>
+  <p>**Example:** The applicability of a rule with a DOM attribute as a test target</a></p>
+  <blockquote>
+    <p>Any `role` attribute that is specified on an HTML or SVG element.</p>
+  </blockquote>
+</div>
 
 ### Applicability for Composite Rules ### {#applicability-composite}
 


act-rules-format.bs
```diff
@@ -296,6 +296,19 @@ Even concepts like headings and images can be misunderstood. These terms could r
   </blockquote>
 </div>
 
+<div class=example>
+  <p>**Example:** The applicability of a rule with the page as a test target</a></p>
+  <blockquote>
+    <p>The rule applies to any page where the root element is an `html` element, and which is not embedded in another page.</p>
```
annethyme: I find it unclear whether the "which" refers to "any page" or "the root element". I think it needs a bit of clarification.

kasperisager: While both "root" and "element" are defined in DOM (https://www.w3.org/TR/dom/#concept-tree-root and https://www.w3.org/TR/dom/#concept-element), I don't think we can use "root element" unambiguously. The term "document element" seems like a better fit: https://www.w3.org/TR/dom/#document-element

WilcoFiers:
@kasperisager @annethyme comments resolved. Please look again.

act-rules-format.bs
```diff
@@ -296,6 +296,19 @@ Even concepts like headings and images can be misunderstood. These terms could r
   </blockquote>
 </div>
 
+<div class=example>
+  <p>**Example:** The applicability of a rule with the page as a test target</a></p>
+  <blockquote>
+    <p>The rule applies to any page where the document element is an `html` element not embedded in another page, for example as part of an `iframe` or `object`.</p>
```
annethyme: Is it the `document element` that is not to be embedded in another page? 
That is how I read the sentence. I don't know if that is technically correct though. I hope you developer guys do know :)

act-rules-format.bs
```diff
@@ -296,6 +296,19 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+  </blockquote>
+</div>
+
+<div class=example>
+  <p>**Example:** The applicability of a rule with a DOM attribute as a test target</a></p>
+  <blockquote>
+    <p>Any `role` attribute that is specified on an HTML or SVG element.</p>
```
annethyme: I suggest changing 

"Any `role` attribute that is specified on an HTML or SVG element."

-->

"The rule applies to any any `role` attribute that is specified on an HTML or SVG element."

... to align better with the other example.

annethyme: @WilcoFiers, is "HTML element" and "SVG element" unambiguous enough for this example in light of our discussion yesterday on whether HTML element meant "element in the HTML namespace", "listed in the HTML specification" or "has its semantics defined in the HTML specification"?

act-rules-format.bs
rdeltour: > where the document element is an `html` element not embedded in another page

this sounds a bit confusing to me: sometimes a page can be both standalone and embedded in a (possibly third-party) iframe…
Maybe we could refer to [WCAG's _Web Page_ definition](https://www.w3.org/TR/WCAG21/#dfn-web-page-s) instead?

rdeltour: or alternatively we could use HTML's _browsing context_ terminology (see [_active document_](https://html.spec.whatwg.org/#active-document) and [_top-level browsing context_](https://html.spec.whatwg.org/#top-level-browsing-context)), but this may gets too technical.

nitedog: Suggestion:

> The rule applies to any page where the document element is an `html` element, and the `html` element is not embedded in another page. Examples of an `html` element embedded in another page can be as part of an `iframe` or `object` elements.

Issue #378, opened 2019-06-13
Update example references from auto-wcag to ACT-R


nitedog:

nitedog:
From ACT TF teleconference call on 13 June 2019: also include the context inside the spec for completeness

Issue #366, opened 2019-05-25
Level Access Feedback: General


WilcoFiers:
**Suggestion**: We need to tighten the outcome data format – possibly restricting it to one form.  The examples suggested for data-format are not trimmed to only those pieces of the format which are 100% necessary.  For example, we likely need the id of the coded test being run, the testing engine name + version, the id of the rule being claimed as supported, the id of the unit test and the outcome from the coded test.

**Reason**: People are already starting to implement tools which read this outcome format.  It would be best to promote a single format ASAP, which of course could be a cut-down form of EARL.   You’ll notice, that both Alfa and aXe already report using different structural versions of EARL (and, Deque’s earl context seems different to the earl-act.json context) – with differences in the additional information presented too.  Without clarity more variants from more tool developers will follow.

WilcoFiers:
@nitedog will add another example to the Appendix.

nitedog:
**Proposed response**

Many thanks for your comment. As discussed during the [ACT TF teleconference call on 27 June 2019](https://www.w3.org/2019/06/27-wcag-act-minutes.html#item11), we believe Pull Request #389 addresses your concern. It provides clearer examples, includes the JSON context, and links out to further resources on EARL and JSON-LD. Please let us know if you disagree with how this was resolved.

aglevelaccess:
Proposed response acceptable.

PR #389, opened 2019-06-27
update Appendix 1

nitedog:
- Removed references to Auto-WCAG and added context, to address issue #378 
- Refined examples, descriptions, and link to further information on EARL/JSON, to address issue #366

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/pull/389.html" title="Last updated on Jul 8, 2019, 4:24 PM UTC (8073320)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wcag-act/389/aac0a4d...8073320.html" title="Last updated on Jul 8, 2019, 4:24 PM UTC (8073320)">Diff</a>

### DIFF
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -695,26 +695,29 @@ Examples in this section include:
   "@context": "context.json",
   "@type": "Assertion",
   "assertedBy": {
-    "@type": "foaf:Organization",
-    "@id": "_:myOrg"
-  }
+    "@type": "Organization",
+    "@id": "_:myOrg",
+    "title": "My Organization",
+    "description" : "Accessibility testing service",
+    "homepage" : "http://example.org/myOrg/"
+  },
   "subject": {
     "@type": ["WebSite", "TestSubject"],
     "@id": "https://example.org/"
-  }
+  },
   "test": {
     "@type": "earl:TestRequirement",
     "@id": "http://www.w3.org/WAI/WCAG2A-Conformance"
   },
   "result": {
-    "outcome": "Failed",
+    "outcome": "earl:failed",
     "source": [{
       "test": {
         "@type": "earl:TestRequirement",
         "@id": "WCAG21:non-text-content"
       },
       "result": {
-        "outcome": "Failed",
+        "outcome": "earl:failed",
         "source": [ … ]
       }
     }, {
@@ -723,7 +726,7 @@ Examples in this section include:
         "@id": "WCAG21:audio-only-and-video-only-prerecorded"
       },
       "result": {
-        "outcome": "Passed",
+        "outcome": "earl:passed",
         "source": [ … ]
       }
     }, {
@@ -732,15 +735,10 @@ Examples in this section include:
         "@id": "WCAG21:captions-prerecorded"
       },
       "result" : {
-        "outcome": "Inapplicable",
+        "outcome": "earl:passed",
         "source": [ … ]
       }
     }, … ]
-  },
-  "myOrg": {
-    "title": "My Organization",
-    "description" : "Accessibility testing service",
-    "homepage" : "http://example.org/myOrg/"
   }
 }
 ```
@@ -754,7 +752,7 @@ Examples in this section include:
   "cnt": "http://www.w3.org/2011/content#",
   "dct": "http://purl.org/dc/terms/",
   "earl": "http://www.w3.org/ns/earl#",
-  "foaf": "http://xmlns.com/foaf/spec/",
+  "foaf": "http://xmlns.com/foaf/0.1/",
   "htp": "http://www.w3.org/2011/http#",
   "ptr": "https://www.w3.org/2009/pointers#",
   "schema": "http://schema.org/",
@@ -775,37 +773,28 @@ Examples in this section include:
   "hasPart": "dct:hasPart",
   "source": "dct:source",
 
-  "Agent": "foaf:term_Agent",
-  "Person": "foaf:term_Person",
-  "Organization": "foaf:term_Organization",
-  "Group": "foaf:term_Group",
-  "Document": "foaf:term_Document",
-  "name": "foaf:term_name",
-  "firstName": "foaf:term_firstName",
-  "surname": "foaf:term_surname",
-  "mbox": "foaf:term_mbox",
-  "mbox_sha1sum": "foaf:term_mbox_sha1sum",
-  "member": "foaf:term_member",
-  "homepage": "foaf:term_homepage",
+  "Agent": "foaf:Agent",
+  "Person": "foaf:Person",
+  "Organization": "foaf:Organization",
+  "Group": "foaf:Group",
+  "Document": "foaf:Document",
+  "name": "foaf:name",
+  "firstName": "foaf:firstName",
+  "surname": "foaf:surname",
+  "mbox": "foaf:mbox",
+  "mbox_sha1sum": "foaf:mbox_sha1sum",
+  "member": "foaf:member",
+  "homepage": "foaf:homepage",
 
-  "assertedBy": { "@id": "@type" },
-  "subject": {
-    "@id": "earl:subject",
-    "@type": "schema:WebPage"
-  },
-  "test": {
-    "@id": "earl:test",
-    "@type": "earl:TestCase"
-  },
-  "result": { "@id": "@type" },
-  "mode": { "@id": "@type" },
-  "outcome": { "@id": "@type" },
+  "assertedBy": { "@type": "@id" },
+  "subject": { "@type": "@id" },
+  "test": { "@type": "@id" },
+  "mode": { "@type": "@id" },
+  "outcome": { "@type": "@id" },
   "pointer": {
     "@id": "earl:pointer",
     "@type": "ptr:CSSSelectorPointer"
-  },
-  "passed": { "@id": "@type" },
-  "failed": { "@id": "@type" }
+  }
 }
 ```
 
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -593,7 +593,7 @@ Appendix 1: Expressing ACT Rule results with JSON-LD and EARL {#appendix-data-ex
 
 This section is *non-normative*.
 
-This section provides examples of expressing results from carrying out ACT Rules using EARL and JSON-LD (See [Evaluation and Report Language](https://www.w3.org/WAI/standards-guidelines/earl/) and [Java Script Object Notation (JSON)](https://www.json.org)). More examples and background is proivided on the [JSON-LD serialization of EARL](https://github.com/w3c/earl) GitHub repository.
+This section provides examples of expressing results from carrying out ACT Rules using EARL and JSON-LD (See [Evaluation and Report Language](https://www.w3.org/WAI/standards-guidelines/earl/) and [A JSON-based Serialization for Linked Data (JSON-LD)](https://www.w3.org/TR/json-ld/)). More examples and background is proivided on the [JSON-LD serialization of EARL](https://github.com/w3c/earl) GitHub repository.
 
 Examples in this section include:
 
--- a/act-rules-format.bs
+++ b/act-rules-format.bs
@@ -609,12 +609,18 @@ Examples in this section include:
 {
   "@context": "context.json",
   "@type": "Assertion",
-  "assertor": "MyTool",
+  "assertedBy": "MyTool",
   "subject": "https://example.org/",
-  "test": "ACT-CG:rules/23a2a8",
+  "test": {
+    "@type": "earl:TestCase",
+    "@value": "ACT-CG:rules/23a2a8"
+  },
   "result": {
     "outcome": "earl:failed",
-    "pointer": "html > body > h1:first-child"
+    "pointer": {
+      "@type": "ptr:CSSSelectorPointer",
+      "html > body > h1:first-child"
+    }
   }
 }
 ```
@@ -626,21 +632,33 @@ Examples in this section include:
   "@context": "context.json",
   "@graph": [{
     "@type": "Assertion",
-    "assertor": "MyTool",
+    "assertedBy": "MyTool",
     "subject": "https://example.org/",
-    "test": "ACT-CG:rules/23a2a8",
+    "test": {
+      "@type": "earl:TestCase",
+      "@value": "ACT-CG:rules/23a2a8"
+    },
     "result": {
       "outcome": "earl:failed",
-      "pointer": "html > body > h1:first-child"
+      "pointer": {
+        "@type": "ptr:CSSSelectorPointer",
+        "html > body > h1:first-child"
+      }
     }
   }, {
     "@type": "Assertion",
-    "assertor": "AnotherTool",
+    "assertedBy": "AnotherTool",
     "subject": "https://example.org/",
-    "test": "ACT-CG:rules/23a2a8",
+    "test": {
+      "@type": "earl:TestCase",
+      "@value": "ACT-CG:rules/23a2a8"
+    },
     "result": {
       "outcome": "earl:passed",
-      "pointer": "html > body > h1:nth-child(2)"
+      "pointer": {
+        "@type": "ptr:CSSSelectorPointer",
+        "html > body > h1:nth-child(2)"
+      }
     }
   }]
 }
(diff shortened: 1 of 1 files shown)


act-rules-format.bs
```diff
@@ -593,38 +593,51 @@ Appendix 1: Expressing ACT Rule results with JSON-LD and EARL {#appendix-data-ex
...
-  "@context": "https://raw.githubusercontent.com/w3c/wcag-act/master/earl-act.json",
+  "@context": "context.json",
   "@type": "Assertion",
+  "assertor": "MyTool",
   "subject": "https://example.org/",
-  "test": "auto-wcag:rules/SC1-1-1-css-image.html",
+  "test": "ACT-CG:rules/23a2a8",
   "result": {
     "outcome": "earl:failed",
     "pointer": "html > body > h1:first-child"
```
WilcoFiers: You need to replace the `pointer` property in the context (my preference) with:

```json
"pointer": {
  "@id": "earl:pointer",
  "@type": "ptr:CSSSelectorPointer",
}
```

Or you need to change this to:

```json
"pointer": {
  "@type": "ptr:CSSSelectorPointer",
  "html > body > h1:first-child"
}
```

act-rules-format.bs
```diff
@@ -699,6 +714,128 @@ This section is *non-normative*.
...
+    "@type": "ptr:PointerPointer"
+  },
+  "info": {
+    "@id": "earl:info",
+    "@type": "xsd:Literal"
+  },
+  "failed": {
+    "@id": "earl:failed",
+    "@type": "Fail",
+    "title": "Failed"
```
WilcoFiers: This is not valid, please remove the `title` property here.

act-rules-format.bs
```diff
@@ -699,6 +714,128 @@ This section is *non-normative*.
...
+    "@id": "earl:mode",
+    "@type": "TestMode"
+  },
+  "mainAssertor": {
+    "@id": "earl:mainAssertor",
+    "@type": "Assertor"
+  },
+  "outcome": {
+    "@id": "earl:outcome",
+    "@type": "OutcomeValue"
```
WilcoFiers: This doesn't resolve right. This should be:

```json
"outcome": { "@id": "@type" },
```

act-rules-format.bs
```diff
@@ -699,6 +714,128 @@ This section is *non-normative*.
...
+  },
+  "NotApplicable": {
+    "@id": "earl:NotApplicable",
+    "@type": "OutcomeValue"
+  },
+  "NotTested": {
+    "@id": "earl:NotTested",
+    "@type": "OutcomeValue"
+  },
+  "Software": "earl:Software",
```
WilcoFiers: (Last of the earl classes that I think can be removed)

act-rules-format.bs
```diff
@@ -699,6 +714,128 @@ This section is *non-normative*.
...
+  "assertedBy": {
+    "@id": "earl:assertedBy",
+    "@type": "Assertor"
+  },
+  "subject": {
+    "@id": "earl:subject",
+    "@type": "TestSubject"
+  },
+  "test": {
```
WilcoFiers: This isn't going to resolve correctly. Turning this:
```json
"test": "ACT-CG:rules/23a2a8"
```

Into this:

```json
"http://www.w3.org/ns/earl#test": {
  "@type": "http://www.w3.org/ns/earl#TestCriterion",
  "@value": "ACT-CG:rules/23a2a8"
}
```

You don't need to put a type here. Because of the `@vocab` that is redundant. More importantly, if you don't want a value, but to have this actually reference external data, you need to do this:

```json
"test": { "@id": "@type" }
```

This gets compacted to the following:

```json
"http://www.w3.org/ns/earl#test": {
  "@id": "https://act-rules.github.io/rules/23a2a8"
},
```

This doesn't have a type, unfortunately. In JSON-LD, you can only give `@value` a default type. If something has to resolve to an `@id`, it can't have a default type. You either have to make it part of the object, or you have to leave the type off.

act-rules-format.bs
```diff
@@ -699,6 +714,128 @@ This section is *non-normative*.
...
+    "@id": "earl:test",
+    "@type": "TestCriterion"
+  },
+  "result": {
+    "@id": "earl:result",
+    "@type": "TestResult"
+  },
+  "mode": {
```
WilcoFiers: This doesn't resolve right. This should be:

```json
"mode": { "@id": "@type" },
```

act-rules-format.bs
```diff
@@ -699,6 +714,128 @@ This section is *non-normative*.
...
+  "Document": "foaf:term_Document",
+  "name": "foaf:term_name",
+  "firstName": "foaf:term_firstName",
+  "surname": "foaf:term_surname",
+  "mbox": "foaf:term_mbox",
+  "mbox_sha1sum": "foaf:term_mbox_sha1sum",
+  "member": "foaf:term_member",
+  "homepage": "foaf:term_homepage",
+
+  "Assertion": "earl:Assertion",
```
WilcoFiers: I think you should remove all of these `earl:` classes. Because you set `@vocab`, they already resolve to earl by default. Also, giving a class a type doesn't really make sense. Only properties can have a value, and so only properties can have a default type. Classes are types. It looks like you're trying to express inheritance in this, but that's not what `@type` is for.

act-rules-format.bs
```diff
@@ -699,6 +714,128 @@ This section is *non-normative*.
...
+  "earl": "http://www.w3.org/ns/earl#",
+  "foaf": "http://xmlns.com/foaf/spec/",
+  "htp": "http://www.w3.org/2011/http#",
+  "ptr": "https://www.w3.org/2009/pointers#",
+  "schema": "http://schema.org/",
+  "xsd": "https://www.w3.org/2001/XMLSchema#",
+
+  "WCAG20": "https://www.w3.org/TR/WCAG20#",
+  "WCAG21": "https://www.w3.org/TR/WCAG21#",
+  "ACT-CG": "https://act-rules.github.io/",
```
WilcoFiers: Could you rename this namespace to `act-rules-cg`?

act-rules-format.bs
```diff
@@ -593,38 +593,51 @@ Appendix 1: Expressing ACT Rule results with JSON-LD and EARL {#appendix-data-ex
...
-**Example 2:**
+**Example 2:** Results from more than one 'Assertion'
 
 ```javascript
 {
-  "@context": "https://raw.githubusercontent.com/w3c/wcag-act/master/earl-act.json",
+  "@context": "context.json",
   "@graph": [{
     "@type": "Assertion",
+    "assertor": "MyTool",
```
WilcoFiers: "assertor" -> "assertedBy"

act-rules-format.bs
```diff
@@ -633,27 +646,28 @@ This section is *non-normative*.
...
-  "test": "wcag21:captions-prerecorded",
+  "test": "WCAG21:captions-prerecorded",
   "result": {
     "outcome": "earl:failed",
     "source": [{
-      "test": "auto-wcag:media-equiv-av-only-alt",
+      "test": "ACT-CG:rules/f51b46",
       "result": {
         "outcome": "earl:failed",
         "source": [ … ]
```
WilcoFiers: It doesn't really make sense for this to have another `source` property. Let's take this out.

act-rules-format.bs
```diff
@@ -633,27 +646,28 @@ This section is *non-normative*.
...
       }
     }, {
-      "test": "auto-wcag:media-equiv-captions",
+      "test": "ACT-CG:rules/fd26cf",
       "result" : {
         "outcome": "earl:failed",
         "source": [ … ]
```
WilcoFiers: Same here. This wouldn't have a source.

act-rules-format.bs
```diff
@@ -663,12 +677,13 @@ This section is *non-normative*.
...
 
-**Example 4:** Aggregate a list of results to a result for the website
+**Example 4:** Aggregating based on 'Test Subject' (eg. for a website)
 
 ```javascript
 {
-  "@context": "https://raw.githubusercontent.com/w3c/wcag-act/master/earl-act.json",
+  "@context": "context.json",
   "@type": "Assertion",
+  "assertor": "Bob",
```
WilcoFiers: "assertor" -> "assertedBy"

act-rules-format.bs
```diff
@@ -633,27 +646,28 @@ This section is *non-normative*.
...
+  "assertor": "MyTool",
   "subject": {
     "@type": ["WebPage", "TestSubject"],
     "@value": "https://example.org/page1.html"
   }
-  "test": "wcag21:captions-prerecorded",
+  "test": "WCAG21:captions-prerecorded",
   "result": {
     "outcome": "earl:failed",
     "source": [{
```
WilcoFiers: Instead of having the source be assertions from two different rules, maybe we should have assertions from the same rule, and add a pointer to each showing they target different elements.

act-rules-format.bs
```diff
@@ -633,27 +646,28 @@ This section is *non-normative*.
...
 
-**Example 3:** Aggregate outcomes to a WCAG success criterion
+**Example 3:** Aggregating based on 'Requirement' (eg. WCAG Success Criteria)
 
 ```javascript
 {
-  "@context": "https://raw.githubusercontent.com/w3c/wcag-act/master/earl-act.json",
+  "@context": "context.json",
   "@type": "Assertion",
+  "assertor": "MyTool",
```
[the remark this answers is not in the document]
nitedog: Done.

act-rules-format.bs
```diff
@@ -593,38 +593,51 @@ Appendix 1: Expressing ACT Rule results with JSON-LD and EARL {#appendix-data-ex
...
+- Example 5: Assumed 'Context' for this section
+
+**Example 1:** Minimal outcome from one 'Assertion'
 
 ```javascript
 {
-  "@context": "https://raw.githubusercontent.com/w3c/wcag-act/master/earl-act.json",
+  "@context": "context.json",
   "@type": "Assertion",
+  "assertor": "MyTool",
```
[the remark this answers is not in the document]
nitedog: Done.

act-rules-format.bs
nitedog: I chose the second approach because the type of pointer should not be controlled by the context but by each assertion separately.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
nitedog: Removed from here because we do not use it in any of the examples.

act-rules-format.bs
nitedog: OK, I changed this but please check -- TestCriterion needs to be refined to TestCase or TestRequirement in each assertion separately.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
```diff
@@ -593,82 +593,127 @@ Appendix 1: Expressing ACT Rule results with JSON-LD and EARL {#appendix-data-ex
...
+    "test": {
+      "@type": "earl:TestCase",
+      "@value": "ACT-CG:rules/23a2a8"
+    },
     "result": {
       "outcome": "earl:failed",
-      "pointer": "html > body > h1:first-child"
+      "pointer": {
+        "@type": "ptr:CSSSelectorPointer",
+        "html > body > h1:first-child"
```
WilcoFiers: This is invalid JSON, you'll have to do `"@value": "html > body > h1:first-child"`.

nitedog: Fixed.

act-rules-format.bs
```diff
@@ -593,82 +593,127 @@ Appendix 1: Expressing ACT Rule results with JSON-LD and EARL {#appendix-data-ex
...
-    "test": "auto-wcag:SC1-1-1-css-image.html",
+    "test": {
+      "@type": "earl:TestCase",
+      "@value": "ACT-CG:rules/23a2a8"
+    },
     "result": {
       "outcome": "earl:passed",
-      "pointer": "html > body > h1:nth-child(2)"
+      "pointer": {
+        "@type": "ptr:CSSSelectorPointer",
```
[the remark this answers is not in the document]
nitedog: If you put in the context then all pointers are of that type. AFAIK you might want to provide different types of pointers depending on the test.

act-rules-format.bs
```diff
@@ -593,82 +593,127 @@ Appendix 1: Expressing ACT Rule results with JSON-LD and EARL {#appendix-data-ex
...
+      }
     }
   }, {
     "@type": "Assertion",
+    "assertedBy": "AnotherTool",
     "subject": "https://example.org/",
-    "test": "auto-wcag:SC1-1-1-css-image.html",
+    "test": {
+      "@type": "earl:TestCase",
+      "@value": "ACT-CG:rules/23a2a8"
```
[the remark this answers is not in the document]
nitedog: Do you mean expand the namespace?

act-rules-format.bs
```diff
@@ -593,82 +593,127 @@ Appendix 1: Expressing ACT Rule results with JSON-LD and EARL {#appendix-data-ex
...
-  "@context": "https://raw.githubusercontent.com/w3c/wcag-act/master/earl-act.json",
+  "@context": "context.json",
   "@graph": [{
     "@type": "Assertion",
+    "assertedBy": "MyTool",
     "subject": "https://example.org/",
-    "test": "auto-wcag:SC1-1-1-css-image.html",
+    "test": {
+      "@type": "earl:TestCase",
+      "@value": "ACT-CG:rules/23a2a8"
```
WilcoFiers: Also, I'm really not sure about expanding all of these objects just so you can give them a type. If your context is `"test": { "@type": "@id" }`, you can do `"test": "ACT-CG:rules/23a2a8"`. It's much easier to read. There is very little value in specifying the type for these.

act-rules-format.bs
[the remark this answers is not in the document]
nitedog: Put into context and added example of overriding that default type.

act-rules-format.bs
nitedog: Done.

act-rules-format.bs
```diff
@@ -732,15 +735,10 @@ Examples in this section include:
         "@id": "WCAG21:captions-prerecorded"
       },
       "result" : {
-        "outcome": "Inapplicable",
+        "outcome": "earl:passed",
         "source": [ … ]
       }
     }, … ]
-  },
-  "myOrg": {
```
WilcoFiers: This wasn't valid.

act-rules-format.bs
```diff
@@ -732,15 +735,10 @@ Examples in this section include:
         "@id": "WCAG21:captions-prerecorded"
       },
       "result" : {
-        "outcome": "Inapplicable",
+        "outcome": "earl:passed",
```
WilcoFiers: A WCAG success criterion isn't allowed to be inapplicable.

act-rules-format.bs
```diff
@@ -695,26 +695,29 @@ Examples in this section include:
...
-    "outcome": "Failed",
+    "outcome": "earl:failed",
     "source": [{
       "test": {
         "@type": "earl:TestRequirement",
         "@id": "WCAG21:non-text-content"
       },
       "result": {
-        "outcome": "Failed",
+        "outcome": "earl:failed",
```
WilcoFiers: These need to be IDs, without a namespace they don't map. It's a quirky thing in JSON-LD.

act-rules-format.bs
```diff
@@ -695,26 +695,29 @@ Examples in this section include:
   "@context": "context.json",
   "@type": "Assertion",
   "assertedBy": {
-    "@type": "foaf:Organization",
-    "@id": "_:myOrg"
-  }
+    "@type": "Organization",
+    "@id": "_:myOrg",
+    "title": "My Organization",
```
WilcoFiers: Moved the organisation in, because the way it was added is incorrect. You could have created a `@graph` array with both the Assertion and the Organisation in as an alternative.

act-rules-format.bs
```diff
@@ -754,7 +752,7 @@ Examples in this section include:
   "cnt": "http://www.w3.org/2011/content#",
   "dct": "http://purl.org/dc/terms/",
   "earl": "http://www.w3.org/ns/earl#",
-  "foaf": "http://xmlns.com/foaf/spec/",
+  "foaf": "http://xmlns.com/foaf/0.1/",
```
WilcoFiers: This is wrong in EARL. I put in the correct terms.

act-rules-format.bs
```diff
@@ -775,37 +773,28 @@ Examples in this section include:
   "hasPart": "dct:hasPart",
   "source": "dct:source",
 
-  "Agent": "foaf:term_Agent",
```
WilcoFiers: None of these need to be prefaced with `term_`. Have a look at this FOAF example. It shows the correct namespace and term to use: http://xmlns.com/foaf/spec/#sec-example

act-rules-format.bs
```diff
@@ -775,37 +773,28 @@ Examples in this section include:
...
-  "Document": "foaf:term_Document",
-  "name": "foaf:term_name",
-  "firstName": "foaf:term_firstName",
-  "surname": "foaf:term_surname",
-  "mbox": "foaf:term_mbox",
-  "mbox_sha1sum": "foaf:term_mbox_sha1sum",
-  "member": "foaf:term_member",
-  "homepage": "foaf:term_homepage",
-
-  "assertedBy": { "@id": "@type" },
```
WilcoFiers: This might have been my mistake, but this needs to be  `{ "@type": "@id" }`. What we're saying here is that when assertedBy is a string instead of an object, the string will be an identifying URL for that object.

act-rules-format.bs
```diff
@@ -775,37 +773,28 @@ Examples in this section include:
...
+  "subject": { "@type": "@id" },
+  "test": { "@type": "@id" },
+  "mode": { "@type": "@id" },
+  "outcome": { "@type": "@id" },
   "pointer": {
     "@id": "earl:pointer",
     "@type": "ptr:CSSSelectorPointer"
-  },
-  "passed": { "@id": "@type" },
-  "failed": { "@id": "@type" }
```
WilcoFiers: passed / failed aren't properties, they are instances. They can only ever be used as a value, so they don't need default type.

PR #398, opened 2019-07-16
Delete static text.

bruce-usab:
It seems crazy to me for a W3C doc to have a link as plain text.  If this edit is not accepted, please, at the very least, change “https://www.w3.org/TR/” to just “http://w3.org/tr”.


annethyme:
Might be a bit weird, but it is completely in line with how it is done in WCAG 2.1: https://www.w3.org/TR/WCAG21/#sotd

nitedog:
Yes, this is boilerplate text for W3C publications, which we cannot remove.

bruce-usab:
> Yes, this is boilerplate text for W3C publications, which we cannot remove.

Is there an editorial party I can comment to?  It makes some sense to have a few URLs for the people looking at the document in print.  But if when one does that, it is only sensible to have the URL as short as possible, even if it is only a few characters.  http://w3.org/tr (16-17 chars, depending on keyboard) is nicer for humans than https://www.w3.org/TR/ (24-25 chars, because of the caps).

nitedog:
I need to check with @michael-n-cooper

michael-n-cooper:
This is boilerplate text that's been around for 20 years and is in thousands of documents. I expect there would be considerable resistance to changing it, and I don't think the issue is worth the effort it would take to even find out where to start.

bruce-usab:
Then can we at least make the text a link as it was with WCAG 2.0?
http://www.w3.org/TR/2008/REC-WCAG20-20081211/Overview.html#status

nitedog:
@bruce-usab I don't understand the difference between the text you refer to and the text you want to edit? (with the exception of https instead of http)

bruce-usab:
1. drop the s from https (1 char)
2. drop the  www from in front of w3.org (4 chars)
3. lowercase tr (saves 2 keystokes)
4. drop the slash at the end (1 char)

nitedog:
@bruce-usab are you aware that you can simply click the link before the URL, which will take you to that same URL? I think the URL was included both as a link and in text back when the argument "what if I need to print the document" was more present.

If you agree, I would like to close the issue here to avoid holding up the publication, and will raise it with our W3C Communications team who I think maintains this front-matter for all W3C specs?

bruce-usab:
> @bruce-usab are you aware that you can simply click the link before the URL

Yes, but that was not my point.  It is retrograde to have something that looks like a URL to not actually be a link.

> I think the URL was included both as a link and in text back when the argument "what if I need to print the document" was more present.

I agree.  But the best practice in that case is to use a as short a link as possible.

> I would like to close the issue here to avoid holding up the publication

Yes please.  I agree to whatever.  No harm, no foul!

> ...and will raise it with our W3C Communications team who I think maintains this front-matter for all W3C specs?

Yes please!

nitedog:
Thanks!

PR #399, opened 2019-07-16
Change “common failures” link

bruce-usab:
Link was not specific enough and the link source does not help clarify what is meant by “common failures” (which was a new term of art to me).


nitedog:
Currently the link goes to the section of WCAG 2.1 that describes the different layers of WCAG, and where the term "common failures" is described. We had opted for that because it is in the normative part of the WCAG specs but I think linking directly to the actual Failures makes more sense.

bruce-usab:
@Shadi, are you inclined to accept this edit or not?

I read that sentence three times, and did not know what it was really talking about.  I followed the link, and still did not figure it out until I thought about it a bit more.

Link text should, of course, reflect the page or nearest heading of the target.  Linking to the normative [Layers of Guidance](http://www.w3.org/TR/WCAG21/#wcag-2-layers-of-guidance) is desirable of course, but the exact phrasing used in the paragraph immediately following the link target is “documented common failures”.

If you want to keep the link as-is, please change “common failures” to “documented common failures” in the middle sentence.  The third sentence can stay as-is.  If that one word — documented — had been in the middle sentence from the beginning, I would not have submitted a pull request.  (But then, I would have missed the proivded typo, so maybe it is all for the best?)

nitedog:
Yes, I personally accept this comment. I expect we will discuss it tomorrow with the rest of the ACT TF.

I think an issue is that WCAG does not really define the term "Failures" anywhere. It speaks of "documented common failures" as you note. Is this something you could consider raising as an issue in AGWG, possibly for consideration in WCAG 2.2?

bruce-usab:
> Is this something you could consider raising as an issue in AGWG, possibly for consideration in WCAG 2.2?

Yes.  But I think bigger the difficulty is that we keep having to explain that if a SC is not-applicable, then the SC is satisfied.  But it is not the least bit clear to me where that particular nugget of information belongs!

nitedog:
> Yes. But I think bigger the difficulty is that we keep having to explain that if a SC is not-applicable, then the SC is satisfied. But it is not the least bit clear to me where that particular nugget of information belongs!

Yes, this is a recurring issue for me too!

PR #422, opened 2019-11-18
Update Common Input Aspects URLs

moekraft:
Update URLs to point to TR version.


moekraft:
Updating URLs to point to TR version of Common Input Aspects note for definitions in Intro. Also, re-adding Mary Jo and Shadi as editors.

PR #521, opened 2021-09-29
Add note about language

Jym77:
All rules with Language as input aspect currently have an assumption that the language is understood by the tester (so that, e.g., a tester can asses whether an accessible name is descriptive or not).
Moving this requirement here.


NOTE-act-rules-common-aspects.bs
```diff
@@ -64,4 +64,6 @@ Language {#input-aspects-text}
...
 
-The means by which the language is assessed, whether by a person or a machine, is not of importance as long as the assessment meets the criteria defined in [[wcag2-tech-req#humantestable]] [[WCAG]].
+Rules can only operate on Language if it can be determined, but the means by which the language is assessed, whether by a person or a machine, is not of importance as long as the assessment meets the criteria defined in [[wcag2-tech-req#humantestable]] [[WCAG]].
+
+Some rules cannot operate on a Language which is not sufficiently understood; while some other only need to be able to assess it. For example, a rule checking that an accessible name is descriptive cannot operate on a Language that is not understood; but a rule checking the correctness of a `lang` attribute can operate on a Language that is not understood.
```
carlosapaduarte, line 69: Trying to streamline, by avoiding double negatives (except in the examples).

```suggestion
Assessing some rules requires understanding the Language, while for others identifying the Language is sufficient. For example, a rule checking that an accessible name is descriptive cannot be assessed if the Language is not understood; but a rule checking the correctness of a `lang` attribute can be assessed even if the Language is not understood.
```

PR #540, opened 2023-08-24
Update act-rules-format.bs for secondary req and acc support

kengdoj:
Add Secondary requirements and accessibility support conclusion from https://github.com/act-rules/act-rules.github.io/discussions/2095

### DIFF
--- a/act-rules-format/act-rules-format.bs
+++ b/act-rules-format/act-rules-format.bs
@@ -226,7 +226,7 @@ A secondary accessibility requirement is a requirement that is correlated with t
 
 When the rule is not designed to test the accessibility requirement, or failed outcomes of the rule still require further testing for the accessibility requirement, the rule <em class="rfc2119">may</em> map the accessibility requirement as Secondary. When an ACT rule maps to a Secondary requirement, it <em class="rfc2119">must</em> include an explanation of why that requirement is Secondary in the Background section of the rule.
 
-When the rule is designed to test an accessibility requirement, limitations or differences in [accessibility support](#accessibility-support) shall not be a reason for that requirement to be a secondary accessibility requirement. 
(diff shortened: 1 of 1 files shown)


act-rules-format/act-rules-format.bs
```diff
@@ -226,6 +226,7 @@ A secondary accessibility requirement is a requirement that is correlated with t
 
 When the rule is not designed to test the accessibility requirement, or failed outcomes of the rule still require further testing for the accessibility requirement, the rule <em class="rfc2119">may</em> map the accessibility requirement as Secondary. When an ACT rule maps to a Secondary requirement, it <em class="rfc2119">must</em> include an explanation of why that requirement is Secondary in the Background section of the rule.
 
+When the rule is designed to test an accessibility requirement, limitations or differences in [accessibility support](#accessibility-support) shall not be a reason for that requirement to be a secondary accessibility requirement.
```
tbostic32, line 229: Should we replace "shall" with either "should" or "must"?

daniel-montalvo: +1 to "should"

act-rules-format/act-rules-format.bs
```diff
@@ -226,6 +226,7 @@ A secondary accessibility requirement is a requirement that is correlated with t
...
```
daniel-montalvo, line 230: Not t sure about the word "limitations" here. Does not "differences" express by itself what we mean here?

daniel-montalvo (changes requested):

act-rules-format/act-rules-format.bs
kengdoj, line 229:
```suggestion
When the rule is designed to test an accessibility requirement, limitations or differences in [accessibility support](#accessibility-support) should not be a reason for that requirement to be a secondary accessibility requirement. 
```

daniel-montalvo (approved):

tbostic32 (approved):

PR #555, opened 2023-12-11
Updating subjective applicability. Including examples and new optiona…

tbostic32:
Including most recently discussed updates to the applicability section.

1. Changed objectivity requirement to **should** instead of **must**. 
2. Added several subjective applicability examples, including one "what not to do" example.
3. Added the optional Applicability Type Designation section.

### DIFF
--- a/act-rules-format/act-rules-format.bs
+++ b/act-rules-format/act-rules-format.bs
@@ -405,11 +405,13 @@ The applicability describes what parts of the [=test subject=] are tested.
 
 ### Applicability for Atomic Rules ### {#applicability-atomic}
 
-The applicability section is a required part of an [=atomic rule=]. It <em class="rfc2119">must</em> contain a precise description of the parts of the [=test subject=] to which the rule applies. For example, specific nodes in the DOM [[DOM]] tree, or tags that are incorrectly closed in an HTML [[HTML]] document. These are known as the [=test targets=]. The applicability <em class="rfc2119">must</em> only use information made available through the listed [input aspects](#input-aspects) in the rule. No other information can be used in the applicability. Applicability <em class="rfc2119">must</em> be described objectively, unambiguously and in plain language.
+The applicability section is a required part of an [=atomic rule=]. It <em class="rfc2119">must</em> contain a precise description of the parts of the [=test subject=] to which the rule applies. For example, specific nodes in the DOM [[DOM]] tree, or tags that are incorrectly closed in an HTML [[HTML]] document. These are known as the [=test targets=]. The applicability <em class="rfc2119">must</em> only use information made available through the listed [input aspects](#input-aspects) in the rule. No other information can be used in the applicability. 
 
-An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, the distance between two elements, etc. Subjective properties on the other hand, are concepts like decorative, navigation mechanism and pre-recorded.
+Applicability <em class="rfc2119">should</em> be described objectively, unambiguously and in plain language. An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, the distance between two elements, etc. When possible, objective descriptions <em class="rfc2119">should</em> be used since they reduce uncertainty. 
 
-Even concepts like headings and images can be misunderstood. These terms could refer to the tag name, the semantic role, or the element's purpose on the web page because they are ambiguous. The latter of which is almost impossible to define objectively. When used in applicability, potentially ambiguous concepts <em class="rfc2119">must</em> be defined objectively. Definitions can be put in the rule [glossary](#glossary), or they can be defined in the section where they are used.
+Subjective properties on the other hand, are concepts like decorative, navigation mechanism and pre-recorded. When possible, subjectivity <em class="rfc2119">should</em> be avoided since it can easily be misunderstood. For example, concepts like headings and images could refer to the tag name, the semantic role, or the element’s purpose on the web page because they are ambiguous. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable.
+
+Definitions can be put in the rule [glossary](#glossary), or they can be defined in the section where they are used.
 
 <aside class=example>
   <header>Example applicability of an atomic rule testing [WCAG 2.1 success criterion 1.4.2 Audio Control](https://www.w3.org/WAI/WCAG21/quickref/#audio-control):</header>
@@ -433,6 +435,32 @@ Even concepts like headings and images can be misunderstood. These terms could r
   </blockquote>
 </aside>
 
+<aside class=example>
+  <header>Example subjective applicability for a rule testing that elements styled as a heading use correct heading markup. This rule applicability cannot be written objectively since "styled as a heading" is subjective and depends on the context of the page.</header>
+  <blockquote>
+    <p>The rule applies to any HTML element that is styled as a heading.</p>
+  </blockquote>
+</aside>
+
+<aside class=example>
+  <header>Example subjective applicability for a rule testing that error messages correctly describe an invalid form field value. THis rule applicability cannot be written objectively since what constitutes a "form field error indicator" is subjective.</header>
+  <blockquote>
+    <p>This rule applies to each HTML element that has one of the following semantic roles: checkbox, combobox, listbox, etc. ..., where the element has a form field error indicator.</p>
+  </blockquote>
+</aside>
+
+<aside class=example>
+  <header>While an objective applicability should be used when possible, care should be taken to ensure parts of the applicability are not included in other sections of the rules format such as the expectation. Below is an <b>incorrect</b> example of using an objective applicability, where the phrase ".... except if the test target is part of a text node that is purely decorative or does not express anything in human language" in the expectation should be included in the applicability.</header>
+  <blockquote>
+    <p>Applicability: This rule applies to any visible character in a text node that is a child in the flat tree of an HTML element </p>
+    <p>Expectation: For each test target, the highest possible contrast between the foreground colors and background colors is at least 3.0:1 for large scale text and 4.5:1 for other texts, except if the test target is part of a text node that is purely decorative or does not express anything in human language.</p>
+  </blockquote>
+</aside>
+
+#### Applicability Type Designation (optional)
+
+Rules can optionally include an applicability type identifier signifying whether the rule contains an objective or a subjective applicability. This identifier is intended to benefit rule readers and implementers by clearly stating the rule authors intention of the applicability and reducing confusion due to different reader and implementer interpretations.
+
 ### Applicability for Composite Rules ### {#applicability-composite}
 
 The applicability of a composite rule is defined as the union of all applicability definitions from the rules listed in the [input rules](#input-rules). Rule authors <em class="rfc2119">may</em> omit a description of the applicability for composite rules. This can be useful if it is difficult to express the combined applicability in plain language. If the composite rule includes applicability, it <em class="rfc2119">must</em> be the union of all the applicability in the [input rules](#input-rules).
@@ -451,6 +479,10 @@ Note that input rules in a composite rule <em class="rfc2119">may</em> have diff
     <p>All `img` elements.</p>
   </blockquote>
 </aside>
+
+#### Applicability Type Designation (optional)
+
+Composite rules can optionally include an applicability type identifier signifying whether the rule contains an objective or a subjective applicability. The applicability type of a composite rule is calculated as subjective if any of the input rules contain a subjective applicability or objective otherwise.
 
 
 Expectations {#expectations}


act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,13 @@ The applicability describes what parts of the [=test subject=] are tested.
...
 
-Even concepts like headings and images can be misunderstood. These terms could refer to the tag name, the semantic role, or the element's purpose on the web page because they are ambiguous. The latter of which is almost impossible to define objectively. When used in applicability, potentially ambiguous concepts <em class="rfc2119">must</em> be defined objectively. Definitions can be put in the rule [glossary](#glossary), or they can be defined in the section where they are used.
+Subjective properties on the other hand, are concepts like decorative, navigation mechanism and pre-recorded. When possible, subjectivity <em class="rfc2119">should</em> be avoided since it can easily be misunderstood. For example, concepts like headings and images could refer to the tag name, the semantic role, or the element’s purpose on the web page because they are ambiguous. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable.
+
+Definitions can be put in the rule [glossary](#glossary), or they can be defined in the section where they are used.
```
tbostic32, line 414: This sentence feels out of place, honestly I don't where it is in the 1.0 format version since it is more of a global statement and not specific to the applicability.

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +435,32 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+</aside>
+
+<aside class=example>
+  <header>While an objective applicability should be used when possible, care should be taken to ensure parts of the applicability are not included in other sections of the rules format such as the expectation. Below is an <b>incorrect</b> example of using an objective applicability, where the phrase ".... except if the test target is part of a text node that is purely decorative or does not express anything in human language" in the expectation should be included in the applicability.</header>
+  <blockquote>
+    <p>Applicability: This rule applies to any visible character in a text node that is a child in the flat tree of an HTML element </p>
+    <p>Expectation: For each test target, the highest possible contrast between the foreground colors and background colors is at least 3.0:1 for large scale text and 4.5:1 for other texts, except if the test target is part of a text node that is purely decorative or does not express anything in human language.</p>
+  </blockquote>
+</aside>
+
```
tbostic32, line 459: Where is the best place to add some informational content on how/where to define "styled as a heading" and "form field error indicator"? Also what is the markup to do an informational section (or is it just add (informational)) to the end of the section header?

act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,13 @@ The applicability describes what parts of the [=test subject=] are tested.
...
 
-An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, the distance between two elements, etc. Subjective properties on the other hand, are concepts like decorative, navigation mechanism and pre-recorded.
+Applicability <em class="rfc2119">should</em> be described objectively, unambiguously and in plain language. An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, the distance between two elements, etc. When possible, objective descriptions <em class="rfc2119">should</em> be used since they reduce uncertainty.
```
WilcoFiers, line 410: MUST be unambiguous, SHOULD be objective and in plain language

act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,13 @@ The applicability describes what parts of the [=test subject=] are tested.
...
 
-Even concepts like headings and images can be misunderstood. These terms could refer to the tag name, the semantic role, or the element's purpose on the web page because they are ambiguous. The latter of which is almost impossible to define objectively. When used in applicability, potentially ambiguous concepts <em class="rfc2119">must</em> be defined objectively. Definitions can be put in the rule [glossary](#glossary), or they can be defined in the section where they are used.
+Subjective properties on the other hand, are concepts like decorative, navigation mechanism and pre-recorded. When possible, subjectivity <em class="rfc2119">should</em> be avoided since it can easily be misunderstood. For example, concepts like headings and images could refer to the tag name, the semantic role, or the element’s purpose on the web page because they are ambiguous. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable.
```
WilcoFiers, line 412:
```suggestion
Subjective properties on the other hand, are concepts like decorative, navigation mechanism and pre-recorded. When possible avoid subjectivity, since it can easily be misunderstood. For example, concepts like headings and images could refer to the tag name, the semantic role, or the element’s purpose on the web page because they are ambiguous. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable.
```

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +435,32 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+  <header>Example subjective applicability for a rule testing that elements styled as a heading use correct heading markup. This rule applicability cannot be written objectively since "styled as a heading" is subjective and depends on the context of the page.</header>
+  <blockquote>
+    <p>The rule applies to any HTML element that is styled as a heading.</p>
+  </blockquote>
+</aside>
+
+<aside class=example>
+  <header>Example subjective applicability for a rule testing that error messages correctly describe an invalid form field value. THis rule applicability cannot be written objectively since what constitutes a "form field error indicator" is subjective.</header>
+  <blockquote>
+    <p>This rule applies to each HTML element that has one of the following semantic roles: checkbox, combobox, listbox, etc. ..., where the element has a form field error indicator.</p>
```
WilcoFiers, line 448: May want to flip this around, make it select the error indicator, not the controls.

act-rules-format/act-rules-format.bs
```diff
@@ -452,6 +480,10 @@ Note that input rules in a composite rule <em class="rfc2119">may</em> have diff
   </blockquote>
 </aside>
 
+#### Applicability Type Designation (optional)
```
WilcoFiers, line 483: I'd like to see this concept added to glossary terms and expectations as well

tbostic32:
I have a secondary PR for adding the subjectivity type to the expectation and the glossary on my own fork if you would like to see it: https://github.com/tbostic32/wcag-act/pull/1. Once this PR is merged I will make a PR for that branch into the W3C/wcag-act main.

act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,13 @@ The applicability describes what parts of the [=test subject=] are tested.
...
 
-Even concepts like headings and images can be misunderstood. These terms could refer to the tag name, the semantic role, or the element's purpose on the web page because they are ambiguous. The latter of which is almost impossible to define objectively. When used in applicability, potentially ambiguous concepts <em class="rfc2119">must</em> be defined objectively. Definitions can be put in the rule [glossary](#glossary), or they can be defined in the section where they are used.
+Subjective properties on the other hand, are concepts like decorative, navigation mechanism and pre-recorded. When possible, avoid subjectivity since it can easily be misunderstood. For example, concepts like headings and images could refer to the tag name, the semantic role, or the element’s purpose on the web page because they are ambiguous. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable.
```
WilcoFiers, line 412: The examples are of ambiguity, not subjectivity.

tbostic32, line 412: In the third sentence, we have
> ... because they are ambiguous.

which basically makes me equate subjectivity to ambiguity (subjectivity == ambiguous). Could we change the wording to differentiate these concepts more? This affects the requirement and the examples below.

WilcoFiers: Maybe this needs have something like: Where possible, a objective parts of a subjective definition can be split into separate rules. For example testing headings with semantic markup separate from headings without semantic markup.

tbostic32: - We need to add examples that are *only* subjective. Steal examples from below or from discussion on act-rules repo. 
- Add "best practice" language for splitting up objective and subjective parts. Required by our usual atomic rule structure, so we don't need to make it a requirement here.

act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,13 @@ The applicability describes what parts of the [=test subject=] are tested.
...
 
-An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, the distance between two elements, etc. Subjective properties on the other hand, are concepts like decorative, navigation mechanism and pre-recorded.
+Applicability <em class="rfc2119">must</em> be unambiguous and <em class="rfc2119">should</em> be described objectively and in plain language. An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, the distance between two elements, etc.
```
tbostic32, line 410: Possibly split this sentence apart and add examples in between the requirements to keep things more localized.

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +435,42 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+  <header>Example subjective applicability for a rule testing that error messages correctly describe an invalid form field value. This rule applicability cannot be written objectively since what constitutes a "form field error indicator" is subjective.</header>
+  <blockquote>
+    <p>Applicability: This rule applies to all form field error indicators.</p>
+    <p>Expectation 1: This form field error indicator allows the identification of the related form field.</p>
+    <p>Expectation 2: The form field error indicator describes the cause of the error or how to solve it.</p>
+  </blockquote>
+</aside>
+
+<aside class=example>
+  <header><strong>Incorrect Example:</strong>This example demonstrates an applicability that attempts to be subjective, but is so broad that it does not meet the "must be unambiguous" requirement. For example, would a text emoji like ":)" be considered non-text content that requires an accessible name? Due to the ambiguity in the applicability it is impossible to tell.</header>
```
tbostic32, line 455: Possibly include that a rule with this applicability is not an atomic rule and should be disallowed.

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +435,42 @@ Even concepts like headings and images can be misunderstood. These terms could r
   </blockquote>
 </aside>
 
+<aside class=example>
+  <header>Example subjective applicability for a rule testing that elements styled as a heading use correct heading markup. This rule applicability cannot be written objectively since "styled as a heading" is subjective and depends on the context of the page.</header>
+  <blockquote>
+    <p>The rule applies to any HTML element that is styled as a heading.</p>
```
tbostic32, line 441: Include definition inline to indicate these are defined terms, otherwise it does seem too broad. 
> Styled as a heading:

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +435,42 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+</aside>
+
+<aside class=example>
+  <header>Example subjective applicability for a rule testing that error messages correctly describe an invalid form field value. This rule applicability cannot be written objectively since what constitutes a "form field error indicator" is subjective.</header>
+  <blockquote>
+    <p>Applicability: This rule applies to all form field error indicators.</p>
+    <p>Expectation 1: This form field error indicator allows the identification of the related form field.</p>
+    <p>Expectation 2: The form field error indicator describes the cause of the error or how to solve it.</p>
+  </blockquote>
+</aside>
```
tbostic32, line 452: Potentially delete this example.

act-rules-format/act-rules-format.bs
tbostic32: Consider an appendix with more extensive examples such as this where more discussion/commentary is needed to describe why the example is good/bad.

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +435,42 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+<aside class=example>
+  <header><strong>Incorrect Example:</strong>This example demonstrates an applicability that attempts to be subjective, but is so broad that it does not meet the "must be unambiguous" requirement. For example, would a text emoji like ":)" be considered non-text content that requires an accessible name? Due to the ambiguity in the applicability it is impossible to tell.</header>
+  <blockquote>
+    <p>Applicability: This rule applies to any non-text content.</p>
+    <p>Expectation: Each test target has an accessible name.</p>
+  </blockquote>
+</aside>
+
+<aside class=example>
+  <header><strong>Incorrect Example:</strong> While an objective applicability should be used when possible, care should be taken to ensure parts of the applicability are not included in other sections of the rules format such as the expectation. Below is an <strong>incorrect</strong> example of using an objective applicability, where the phrase, ".... except if the test target is part of a text node that is purely decorative or does not express anything in human language" in the expectation should be included in the applicability.</header>
```
tbostic32, line 463: Can be more explicit and give exact correct applicability: 
> This rule applies to any visible character in a text node that is a child in the flat tree of an HTML element, except if the test target is part of a text node that is purely decorative or does not express anything in human language.

act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,15 @@ The applicability describes what parts of the [=test subject=] are tested.
...
+Additionally, the applicability <em class="rfc2119">should</em> be described objectively and in plain language. An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, and the distance between two elements.
+
+Subjective properties, on the other hand, are concepts like decorative, navigation mechanism, and pre-recorded that may be defined a variety of ways, often relying on human judgement and hueristics. When possible, avoid subjectivity since it can easily be misunderstood. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable. When crafting a subjective applicability, if possible, split the objective and subjective parts of the applicaility into separate rules. This will ensure the designed rule meets the atomic rule requirement of testing a single condition. An example would be testing headings with correct semantic markup in a different rule from those without semantic markup.
```
daniel-montalvo, line 414:
```suggestion
Subjective properties, on the other hand, are concepts like decorative, navigation mechanism, and pre-recorded that may be defined a variety of ways, often relying on human judgement and heuristics. When possible, avoid subjectivity since it can easily be misunderstood. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable. When crafting a subjective applicability, if possible, split the objective and subjective parts of the applicability into separate rules. This will ensure the designed rule meets the atomic rule requirement of testing a single condition. An example would be testing headings with correct semantic markup in a different rule from those without semantic markup.
```

act-rules-format/act-rules-format.bs
```diff
@@ -452,15 +454,15 @@ Definitions can be put in the rule [glossary](#glossary), or they can be defined
 </aside>
 
 <aside class=example>
-  <header><strong>Incorrect Example:</strong>This example demonstrates an applicability that attempts to be subjective, but is so broad that it does not meet the "must be unambiguous" requirement. For example, would a text emoji like ":)" be considered non-text content that requires an accessible name? Due to the ambiguity in the applicability it is impossible to tell.</header>
+  <header><strong>Incorrect Example:</strong>This example demonstrates an applicability that attempts to be subjective, but does not meet the "must be unambiguous" requirement. For example, non-text content could refer to images, interactive elements, or even a text emoji like ":)". Due to the ambiguity in the applicability it is impossible to tell.</header>
```
WilcoFiers, line 457: Do we need to add markup to that smiley?

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +437,42 @@ Even concepts like headings and images can be misunderstood. These terms could r
   </blockquote>
 </aside>
 
+<aside class=example>
+  <header>Example subjective applicability for a rule testing that elements styled as a heading use correct heading markup. This rule applicability cannot be written objectively since "styled as a heading" is subjective and depends on the context of the page.</header>
+  <blockquote>
+    <p>The rule applies to any HTML element that is styled as a heading. Elements that are styled as a heading may include features such as a larger font-size than nearby text, bolding or changing of the font, or the usage of whitespace or shapes visually distinguish the element text.</p>
```
daniel-montalvo, line 443:
```suggestion
    <p>The rule applies to any HTML element that is styled as a heading. Elements that are styled as a heading may include features such as a larger font-size than nearby text, bolding or changing of the font, or the use of whitespace or shapes that visually distinguish the element text.</p>
```

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +437,42 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+  <header>Example subjective applicability for a rule testing that error messages correctly describe an invalid form field value. This rule applicability cannot be written objectively since what constitutes a "form field error indicator" is subjective.</header>
+  <blockquote>
+    <p>Applicability: This rule applies to all form field error indicators.</p>
+    <p>Expectation 1: This form field error indicator allows the identification of the related form field.</p>
+    <p>Expectation 2: The form field error indicator describes the cause of the error or how to solve it.</p>
+  </blockquote>
+</aside>
+
+<aside class=example>
+  <header><strong>Incorrect Example:</strong>This example demonstrates an applicability that attempts to be subjective, but does not meet the "must be unambiguous" requirement. For example, non-text content could refer to images, interactive elements, or even a text emoji like ":)". Due to the ambiguity in the applicability it is impossible to tell.</header>
```
daniel-montalvo, line 457:
```suggestion
  <header><strong>non-conforming Example:</strong>This example demonstrates an applicability that attempts to be subjective, but does not meet the "must be unambiguous" requirement. For example, non-text content could refer to images, interactive elements, or even a text emoji like ":)". Due to the ambiguity in the applicability it is impossible to tell.</header>
```

act-rules-format/act-rules-format.bs
tbostic32, line 414:
```suggestion
Subjective properties, on the other hand, are concepts like decorative, navigation mechanism, and pre-recorded that may be defined in a variety of ways, often relying on human judgement and heuristics. When possible, avoid subjectivity since it can easily be misunderstood. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable. When crafting a subjective applicability, if possible, split the objective and subjective parts of the applicability into separate rules. This will ensure the designed rule meets the atomic rule requirement of testing a single condition. An example would be testing headings with correct semantic markup in a different rule from those without semantic markup.
```

act-rules-format/act-rules-format.bs
tbostic32: Add accessible name "Text smiley face created using a colon and parenthesis".

act-rules-format/act-rules-format.bs
tbostic32: Potentially could include in an appendix or note.

act-rules-format/act-rules-format.bs
tbostic32: Need to do this for each example.

act-rules-format/act-rules-format.bs
tbostic32: I will add this in a separate PR after this is merged.

act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,15 @@ The applicability describes what parts of the [=test subject=] are tested.
...
 
-An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, the distance between two elements, etc. Subjective properties on the other hand, are concepts like decorative, navigation mechanism and pre-recorded.
+Applicability <em class="rfc2119">must</em> be unambiguous so that the applicability can only be understood in one way. For example, concepts like headings and images are ambiguous since they could refer to the tag name, the semantic role, or the element’s purpose on the web page. Oppositely, a rule that specifically only applies to elements with a heading tag would satisfy the unambiguous requirement.
```
kengdoj, line 410: "Oppositely" sounds strange. Not sure it's needed. Maybe "conversely" would be better.

act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,15 @@ The applicability describes what parts of the [=test subject=] are tested.
...
 
-Even concepts like headings and images can be misunderstood. These terms could refer to the tag name, the semantic role, or the element's purpose on the web page because they are ambiguous. The latter of which is almost impossible to define objectively. When used in applicability, potentially ambiguous concepts <em class="rfc2119">must</em> be defined objectively. Definitions can be put in the rule [glossary](#glossary), or they can be defined in the section where they are used.
+Additionally, the applicability <em class="rfc2119">should</em> be described objectively and in plain language. An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, and the distance between two elements.
```
kengdoj, line 412: what does "distance between two elements" mean?

act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,15 @@ The applicability describes what parts of the [=test subject=] are tested.
...
+
+Subjective properties, on the other hand, are concepts like decorative, navigation mechanism, and pre-recorded that may be defined in a variety of ways, often relying on human judgement and heuristics. When possible, avoid subjectivity since it can easily be misunderstood. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable. When crafting a subjective applicability, if possible, split the objective and subjective parts of the applicability into separate rules. This will ensure the designed rule meets the atomic rule requirement of testing a single condition. An example would be testing headings with correct semantic markup in a different rule from those without semantic markup.
+
+Definitions can be put in the rule [glossary](#glossary), or they can be defined in the section where they are used.
 
 <aside class=example>
   <header>Example applicability of an atomic rule testing [WCAG 2.1 success criterion 1.4.2 Audio Control](https://www.w3.org/WAI/WCAG21/quickref/#audio-control):</header>
```
kengdoj, line 419:
```suggestion
  <header>Example objective applicability of an atomic rule testing [WCAG 2.1 success criterion 1.4.2 Audio Control](https://www.w3.org/WAI/WCAG21/quickref/#audio-control):</header>
```

(similar suggestion for the following objective examples)

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +437,33 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+  <blockquote>
+    <p>The rule applies to any HTML element that is styled as a heading. Elements that are styled as a heading may include features such as a larger font-size than nearby text, bolding or changing of the font, or the use of whitespace or shapes that visually distinguish the element text.</p>
+  </blockquote>
+</aside>
+
+<aside class=example>
+  <header><strong>Non-conforming Example:</strong>This example demonstrates an applicability that attempts to be subjective, but does not meet the "must be unambiguous" requirement. For example, non-text content could refer to images, interactive elements, or even a text emoji like <span role="figure" aria-label="Text smiley face created using a colon and parenthesis">:)</span>. Due to the ambiguity in the applicability it is impossible to tell.</header>
+  <blockquote>
+    <p>Applicability: This rule applies to any non-text content.</p>
+    <p>Expectation: Each test target has an accessible name.</p>
```
kengdoj, line 451: For consistency across examples, suggest:
- either remove Applicability here or add it to previous examples
- remove the Expectation as other examples don't have it

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +437,33 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+  <blockquote>
+    <p>Applicability: This rule applies to any visible character in a text node that is a child in the flat tree of an HTML element </p>
+    <p>Expectation: For each test target, the highest possible contrast between the foreground colors and background colors is at least 3.0:1 for large scale text and 4.5:1 for other texts, except if the test target is part of a text node that is purely decorative or does not express anything in human language.</p>
+  </blockquote>
+</aside>
+
+#### Applicability Type Designation (optional)
+
+Rules can optionally include an applicability type identifier signifying whether the rule contains an objective or a subjective applicability. This identifier is intended to benefit rule readers and implementers by clearly stating the rule authors intention of the applicability and reducing confusion due to different reader and implementer interpretations.
```
kengdoj, line 465:
```suggestion
Rules can optionally include an applicability type identifier signifying whether the rule contains an objective or a subjective applicability. This identifier is intended to benefit rule readers and implementers by clearly stating the rule author's intention of the applicability and reducing confusion due to different reader and implementer interpretations.
```

act-rules-format/act-rules-format.bs
tbostic32: I considered using conversely, but landed on oppositely since I thought it was a slightly simpler term, lets see what the group thinks.

act-rules-format/act-rules-format.bs
tbostic32: The could be some slight differences in the calculation, but generally you would grab the (x,y) pixel location of two elements on a page and then calculate the distance between those locations.

act-rules-format/act-rules-format.bs
tbostic32:
```suggestion
Applicability <em class="rfc2119">must</em> be unambiguous so that the applicability can only be understood in one way. For example, concepts like headings and images are ambiguous since they could refer to the tag name, the semantic role, or the element’s purpose on the web page. Conversely, a rule that specifically only applies to elements with a heading tag would satisfy the unambiguous requirement. 
```

act-rules-format/act-rules-format.bs
tbostic32: Add "Applicability:" for the other examples, remove expectation here.

act-rules-format/act-rules-format.bs
tbostic32:
```suggestion
Additionally, the applicability <em class="rfc2119">should</em> be described objectively and in plain language. An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, and the calculated distance between two elements.
```

Or should we consider using a different example, e.g., the color value of a pixel.

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +437,33 @@ Even concepts like headings and images can be misunderstood. These terms could r
...
+  <blockquote>
+    <p>Applicability: This rule applies to any visible character in a text node that is a child in the flat tree of an HTML element </p>
+    <p>Expectation: For each test target, the highest possible contrast between the foreground colors and background colors is at least 3.0:1 for large scale text and 4.5:1 for other texts, except if the test target is part of a text node that is purely decorative or does not express anything in human language.</p>
+  </blockquote>
```
Jym77, line 460: **Issue:** From a rather logical/specification point of view, there is nothing wrong with that rule. It is only "wrong" because we know that the corresponding SC doesn't apply to non-text content and therefore the exception should be in the Applicability.
But if we change the Expectation to "either the target has good contrast, or it is non-text content", it suddenly doesn't look that bad. And for some other SC, we have "either the text size can be changed, or size is essential" where the essential size is subjective and but is natural in the Expectation.

From a more abstract point of view, consider:
* Applicability: (something objective), except if (something subjective)
  Expectation: (something objective)
* Applicability: (something objective)
  Expectation: (something objective) or (something subjective)
* Applicability: (something objective)
  Expectation: (something objective), except if (something subjective)

All three are doing the same thing, and depending on the Accessibility Requirement, could be legit. So, I'm uneasy about saying that this example is "bad" based on, well, mostly subjective criteria 😉
(I do agree it is bad and cheating around objectivity by moving part of the Applicability in the Expectation; I do not feel we can really define why it's bad in terms objective enough for a specification).

---
Additionally, the normative part of the text says nothing about "do not cheat by moving stuff in the Expectation", which makes this example come quite out of the blue. The previous example illustrate the "ambiguous" vs "subjective" difference that is mentioned earlier in the text, but this example looks out of place (unless you have the historical context of ACT rules and know we did that trick for a long time while format v1.0 was still a thing).

Jym77, line 460: **Note:** Additionally, this also makes our published "color contrast" rules not matching the format, so it definitely needs a version bump on the format.
(so that the current rules can still be valid as per format 1.0, until we upgrade them to the newer format)

act-rules-format/act-rules-format.bs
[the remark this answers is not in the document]
tbostic32: Yes, I think primarily due to how hard it is to measure if text is in plain language we downgraded the requirement to SHOULD.

act-rules-format/act-rules-format.bs
```diff
@@ -405,11 +405,15 @@ The applicability describes what parts of the [=test subject=] are tested.
...
+Additionally, the applicability <em class="rfc2119">should</em> be described objectively and in plain language. An objective description is one that can be resolved without uncertainty, in a given technology. Examples of objective properties in HTML are tag names, their computed role, and the distance between two elements.
+
+Subjective properties, on the other hand, are concepts like decorative, navigation mechanism, and pre-recorded that may be defined in a variety of ways, often relying on human judgement and heuristics. When possible, avoid subjectivity since it can easily be misunderstood. In cases where it is impossible to objectively define the applicability, the use of a subjective description in the applicability is acceptable. When crafting a subjective applicability, if possible, split the objective and subjective parts of the applicability into separate rules. This will ensure the designed rule meets the atomic rule requirement of testing a single condition. An example would be testing headings with correct semantic markup in a different rule from those without semantic markup.
```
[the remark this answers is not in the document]
tbostic32: I will take it back for discussion with TF. My gut feeling is that we could potentially add it as a Note if we wanted to cover it. We can also emphasize the subjectivity more in the note, we did decide to have multiple places that discourage using subjectivity (both the "SHOULD be objective requirement" and the "When possible, avoid subjectivity...") so hopefully that helps to dissuade.

act-rules-format/act-rules-format.bs
```diff
@@ -433,6 +437,33 @@ Even concepts like headings and images can be misunderstood. These terms could r
   </blockquote>
 </aside>
 
+<aside class=example>
+  <header>Example subjective applicability for a rule testing that elements styled as a heading use correct heading markup. This rule applicability cannot be written objectively since "styled as a heading" is subjective and depends on the context of the page.</header>
+  <blockquote>
+    <p>The rule applies to any HTML element that is styled as a heading. Elements that are styled as a heading may include features such as a larger font-size than nearby text, bolding or changing of the font, or the use of whitespace or shapes that visually distinguish the element text.</p>
```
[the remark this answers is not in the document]
tbostic32: Agreed, I think the hard part would be writing the requirement in a way that doesn't stifle us in the future. Do we always want to have additional criteria - probably. Will those criteria always be objective - probably not (I don't think some the current criteria are actually objective). I will see what others in the TF might think about this.

act-rules-format/act-rules-format.bs
tbostic32: I would disagree that the three abstract forms you show up there result in the same behavior. Specifically, we are trying to prevent cases that should be inapplicable showing up as passed (understanding that passed and inapplicable are often treated similarly for us). When you use form number from your comment:

> Applicability: (something objective), except if (something subjective)
> Expectation: (something objective)

The "except if (something subjective)" clause in the applicability will prevent the rule from be evaluated on content that meets the accepted criteria, thus making it inapplicable. 

> Applicability: (something objective)
> Expectation: (something objective) or (something subjective)

> Applicability: (something objective)
> Expectation: (something objective), except if (something subjective)

However, for these second two forms the rule **WILL** be evaluated because the applicability matches, but the expectation will PASS the rule since it is covered by an exception in the Expectation. With that said, the point still remains if we care if one form says inapplicable while the other says passed - we could have a deeper discussion on that point.

I appreciate the thought that this seems to come out of the blue, we could perhaps frame this up better earlier on in the document. My thought process was to show a type of "anti-pattern" that we were hoping to dissuade the use of in the future as we didn't feel it was following the intentions of the format.

Jym77: Yes, they are not equivalent indeed. They report the same failed targets, which is what the rules mostly care about.

My point is more than having this one as a bad practice only stems from the SC it applies to, not from the shape of the expectation itself. That is, the SC is phrased in a way that non-text is "inapplicable", so we want the rule to reflect it. But using "except (some condition)" or even "except (subjective condition)" in the Expectation is not bad per se.

Staying on 1.4.3, consider:

* Applicability: Any character in any text node (as current)
*  Expectation: good contrast, except if non-text or if logotype.

Now, the first exception (non-text) is "bad" because the SC only applies to text. But the second exception (logotype) is "good" because the SC does apply to (text) logotypes and has a special exception in its requirement.

(actually, we may argue whether logotypes should be passing or inapplicable for such a rule…)

From a "find errors" point of view, whether these exceptions are part of the Applicability or Expectation makes no difference. From a formal/logical point of view, there is no real difference between these exceptions. I'm a bit reluctant in phrasing "that" as an anti-pattern without really being able to define what "that" is. It is "trying to sneak the subjective part of Applicability into the Expectation", but that itself feels quite subjective since I find it hard to decide whether something should be Inapplicable or Passing, even more without the link to the requirement the rule should map to.

So, as much as I like this guideline, and it feels like a good one for developers of rules (e.g., in the CG's design document), I'm also not sure it belongs to the rule format itself.

---
I'm also not sure about "forcing" this guideline upon other groups (than the ACT rules CG) who want to maintain a ruleset (e.g., at Siteimprove we have our own ruleset with copies of most rules from the CG, plus several custom rules). Changes we make to the format de facto force other groups to adapt or stop saying their ruleset is following the format. I believe the format benefits from having several rulesets that can notably be used as incubators for rules to be ultimately approved by the AG and included in Understanding WCAG.

tbostic32: If I could push the conversation to another rule as an example for a moment. The [Error message describes invalid form field value](https://www.w3.org/WAI/standards-guidelines/act/rules/36b590/proposed/) rule uses the following text in each of its three expectations:

> Each test target either has no [form field error indicators](https://www.w3.org/WAI/standards-guidelines/act/rules/36b590/proposed/#form-field-error-indicator), or at least one of the [form field error indicators](https://www.w3.org/WAI/standards-guidelines/act/rules/36b590/proposed/#form-field-error-indicator) ....

I originally had this as another "anti-pattern" example but removed it since it was duplicative and *I thought* more complex than the color contrast rule above. In this case, I think it is pretty clear that the pushing the applicability into the expectation makes the rule significantly more complex and harder to read. 

So maybe the conversation should be more about when including exceptions in the Expectation vs the Applicability makes more sense. Since they may be both subjective at this point both are a possibility. Here are some reasons to disallow one or the other, i use the term "applicability statements" to mean statements that describe what targets should be tested even when those statements may occur in the expectation.

1. Putting applicability statements in the Expectation vs the Applicability affects the failed test targets -> Applicability statement should go where it makes logical sense and results in the intended outcome for the test targets. 
2. Including exceptions in Expectation or Applicability significantly reduces readability/does not meet the plain language requirement -> Applicability should go in the place that best meets the plain language requirements
3. As a default, applicability statements should always go Applicability unless significant reason for placing it in the Expectation can be show **AND** putting an applicability statement in the expectation does not result in a change of failed test targets.  

I'm not sure I am fully convinced, but I am trying to meet you halfway and find some more formal language we could use to describe the situation.

Jym77: I pretty much agree that "Applicability statements" should go in the Applicability unless there is some very good reason to do otherwise. They are **Applicability** statements, not Expectation statements…

My problem resides in the fact that "Applicability statements" is not really a well defined term. It is at the very least related to the Accessibility requirement the rule is designed for. But ultimately, since the rules are focusing on detecting true errors, it matter little whether something is Passed or Inapplicable. It only matters because we link it back to WCAG and agree that "a page with only non-text content is Inapplicable to 1.4.3, not Passing, so it should be Inapplicable for the ACT rule too". (and in the case of the error indicators, the rule would be more compact and readable if the "no error indicator" bit was factored in the Applicability…)

From the point of view of the ruleset (i.e., the ACT rules CG), it is a good design principle to avoid "Applicability statements" in the Expectation, and we can live with a blurry definition of them. From the point of the rule format, I am not sure we can define "Applicability statement" well enough to enforce (MUST) or even recommend (SHOULD) "no Applicability statement in the expectation".

(plus the fact that the "no applicability statement in the expectation" bit is mentioned in that example without being mentioned earlier in the normative text).

---
Taking another example.

A rule for [2.4.1 Bypass Blocks](https://www.w3.org/TR/WCAG22/#bypass-blocks) could be either Passing or Inapplicable on pages without blocks of repeated contents. Our current rule is Passing them. E.g., [Document has heading for non-repeated content](https://www.w3.org/WAI/standards-guidelines/act/rules/047fe0/proposed/) has Expectation "either there is no [non-repeated content after repeated content](https://www.w3.org/WAI/standards-guidelines/act/rules/047fe0/proposed/#non-repeated-content) or …"

I think this is another "Applicability statement" put in the Expectation for the sake of objectivity of the Applicability. Because the SC doesn't really apply to pages without repeated content.

Now, consider the (PR) best practice rule [First focusable element is link to non-repeated content](https://github.com/act-rules/act-rules.github.io/pull/1539). It doesn't map to any SC. It does relate to technique G1, but that technique isn't really restricted to pages with repeated content.

Currently, that PR also passes pages without repeated content (actually quite specifically with one of its atomic rules!) Is that also an "Applicability statement"? This is much less clear. Telling devs "[applicability] pages should [expectation] either have no repeated content¹, or a skip link" is a decent best practice, as much as "[applicability] pages with repeated content should [expectation] have a skip link". I tend to think that "no repeated content" is an "Applicability statement" here too, but only because the rule indirectly links to 2.4.1 and I use the same reasoning. I.e., a much weaker case than for the full blown "bypass block" rules before (cannot be backed directly to some other normative text). And since having "no repeated content" in the Expectation causes no problem (no false positives), I wouldn't fight to have it moved.

¹ I've seen several customers with "single download" pages that wrap a media file without any other content and thus have no repeated content.

So, in that case, "no repeated content" could be an "applicability statement" or not depending on a lot of outer circumstances. That is why I am uncomfortable with pointing out "no applicability statement in expectation" in the rules format specification.

act-rules-format/act-rules-format.bs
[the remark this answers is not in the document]
tbostic32: From TF group, consider adding a NOTE on objectivity intention.

act-rules-format/act-rules-format.bs
```diff
@@ -405,34 +405,64 @@ The applicability describes what parts of the [=test subject=] are tested.
...
+  <blockquote>
+    <p>Applicability: This rule applies to any visible character in a text node that is a child in the flat tree of an HTML element </p>
+    <p>Expectation: For each test target, the highest possible contrast between the foreground colors and background colors is at least 3.0:1 for large scale text and 4.5:1 for other texts, except if the test target is part of a text node that is purely decorative or does not express anything in human language.</p>
+  </blockquote>
+</aside>
+
+#### Applicability Type Designation (optional)
+
+Rules can optionally include an applicability type identifier signifying whether the rule contains an objective or a subjective applicability. This identifier is intended to benefit rule readers and implementers by clearly stating the rule author's intention of the applicability and reducing confusion due to different reader and implementer interpretations.
```
tbostic32, line 464: Should further text be included here to allow authors to state their intention?

act-rules-format/act-rules-format.bs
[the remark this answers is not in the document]
tbostic32: From comment: https://github.com/w3c/wcag-act/pull/555#discussion_r1458885060

Create NOTE or add "Best Practice" language for describing how to resolve subjective applicability cases.

act-rules-format/act-rules-format.bs
```diff
@@ -405,34 +405,64 @@ The applicability describes what parts of the [=test subject=] are tested.
...
+  <blockquote>
+    <p>Applicability: This rule applies to any non-text content.</p>
+  </blockquote>
+</aside>
+
+<aside class=example>
+  <header><strong>Non-conforming Example:</strong> While an objective applicability should be used when possible, care should be taken to ensure parts of the applicability are not included in other sections of the rules format such as the expectation. Below is an <strong>incorrect</strong> example of using an objective applicability, where the phrase, ".... except if the test target is part of a text node that is purely decorative or does not express anything in human language" in the expectation should be included in the applicability. When fixed, the full applicability would be, "This rule applies to any visible character in a text node that is a child in the flat tree of an HTML element, except if the test target is part of a text node that is purely decorative or does not express anything in human language."</header>
```
tbostic32, line 455: Show compare and contrast example between this original example and then the subjective one that is preferred. This is not a non-conforming example.

act-rules-format/act-rules-format.bs
```diff
@@ -405,34 +405,71 @@ The applicability describes what parts of the [=test subject=] are tested.
...
   </blockquote>
 </aside>
 
+<aside class=example>
+  <header>The below section contains 2 approaches for writing an ACT rule for testing text contrast. For each example, both the applicability and expectation are included for clarity. Each of these examples follows the ACT rules format, but note that in the first example, the applicability is ended with the phrase "... except if the test target is part of a text node that is purely decorative or does not express anything in human language", while in the second example this phrase is appended to the expectation. Both of these approaches follow the normative ACT rules format and lead to valid ACT rules; however, we recommend including this text in the applicability when possible. When phrases such as this are included in the applicability, some test cases become inapplicable that would otherwise be passed if the phrase was included in the expectation. For example, for a smiling face emoji would be considered inapplicable when using the approach in Example 1, while in Example 2 the smiling face emoji would pass since it is considered an exception to the expectation.</header>
```
tbostic32, line 455: Move out of this document and into CG writing guidelines.

PR #581, opened 2024-12-06
Update act-rules-format.html

kengdoj:
- remove secondary req explanation in background section
- update cantTell and Untested definitions


daniel-montalvo:
Hi @kengdoj 

thanks for working on these changes.

The act-rules-format/Overview.html file is automatically generated by  the spec processor we are using (bikeshed). We should not manually edit it. I suggest we close this PR and open another one by editing https://github.com/w3c/wcag-act/blob/main/act-rules-format/act-rules-format.bs which is on the main branch.

I will review when I am back.

WilcoFiers (approved):
LGTM

daniel-montalvo (changes requested):
Hi @kengdoj 

thanks for working on these changes.

The act-rules-format/Overview.html file is automatically generated by  the spec processor we are using (bikeshed). We should not manually edit it. I suggest we close this PR and open another one by editing https://github.com/w3c/wcag-act/blob/main/act-rules-format/act-rules-format.bs which is on the main branch.

I will review when I am back.

PR #584, opened 2025-02-24
Determine if a rule identifier is unique

### LINKED ISSUE #572: How are rule identifiers matched to one another?
(This is part of the review by the Internationalization WG. Sorry for being late – it's entirely my fault.)

4.1. Rule Identifier
https://www.w3.org/TR/2024/WD-act-rules-format-1.1-20240618/#rule-identifier

> This identifier must be unique when the rule is part of a ruleset. The identifier can be any text [...]

To know if an identifier is unique (and to be able to use it in one rule to point to another), you need to know when two identifiers are the same. E.g., are capital letters (ABC) the same as lowercase letters (abc)? If a letter can be encoded in Unicode in two ways (e.g., ‘é’ as single character vs separate ‘e’ + acute accent) are those the same?

[‘Character Model for the World Wide Web: String Matching’](https://www.w3.org/TR/charmod-norm/) explains the issues with comparing two strings of text and has recommendations for choosing an algorithm, including for text strings used as identifiers.

daniel-montalvo:
Fixes #572


act-rules-format/act-rules-format.bs
```diff
@@ -132,6 +132,8 @@ Rule Identifier {#rule-identifier}
 
 An ACT Rule <em class="rfc2119">must</em> have an identifier. This identifier <em class="rfc2119">must</em> be unique when the rule is part of a ruleset. The identifier can be any text, such as plain text, URL, or a database identifier.
 
+Rulesets <em>MUST</em> be consistent when determining if an identifier is unique. For example, by using lowercase ASCII or other unambiguous method. For a detailed explanation of the problem, see [‘Character Model for the World Wide Web: String Matching’](https://www.w3.org/TR/charmod-norm/)
```
daniel-montalvo, line 135: Tweak it because there is not such a thing as a "ruleset" in the Format. Maybe "when an individual rule is to be used within a ruleset" or similar.

daniel-montalvo: updated. Let's discuss further in the meeting.

kengdoj (approved):

